PR #160, opened 2026-07-12
fix: return proper HTTP error statuses

### LINKED ISSUE #120: [Fix] Return proper HTTP status codes on validation and auth failures
### Description
Several route handlers return `200 OK` on validation or authorization failures, so clients read failures as success.

### Why
`students/[code]` and `user` routes do `return NextResponse.json({ message: safeParse.error })` with no status (defaults to 200); unauthorized paths also return 200.

### How
Return `400` for validation errors and `401/403` for auth. Fold into a shared response helper as part of the service-layer work.

### Dependencies
None.

dinisjcorreia:
## Summary

- add a shared API error response helper for 400/401/403 responses
- return 400 for validation failures in student, user, and saved PATCH routes
- return 401/403 for student and user authorization failures through the shared helper

Closes #120

## Verification

- `git diff --check`
- focused runtime assertions for helper status and JSON body across 400/401/403
- Prettier passes for the new helper and named issue routes
- ESLint remains blocked by the pre-existing flat-config circular-reference crash
- TypeScript remains blocked by pre-existing config/app errors; no errors reported in the new helper or named issue routes
- `saved/route.ts` retains pre-existing whole-file Prettier findings; task lines are formatted


src/services/apiResponse.ts
```diff
@@ -0,0 +1,7 @@
+import { NextResponse } from "next/server";
+
+type ErrorStatus = 400 | 401 | 403;
+
+export function errorResponse(error: unknown, status: ErrorStatus) {
+  return NextResponse.json({ error }, { status });
+}
```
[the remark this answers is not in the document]
dinisjcorreia: Addressed in eb08728. `errorResponse` now preserves strings, uses `Error.message` for Error/Zod instances, and stringifies other values so arbitrary object fields are not returned. Ran focused runtime checks for strings, Error, ZodError, and arbitrary objects.

src/services/apiResponse.ts
```diff
@@ -0,0 +1,8 @@
+import { NextResponse } from "next/server";
+
+type ErrorStatus = 400 | 401 | 403;
+
+export function errorResponse(error: unknown, status: ErrorStatus) {
+  const message = error instanceof Error ? error.message : String(error);
+  return NextResponse.json({ error: message }, { status });
```
david-fs-valente: Minor: for Zod errors, error.message is already a JSON-stringified array of issues (that's how ZodError builds its .message), so this ends up double-encoding — the response body becomes {"error": "[{\"code\":\"invalid_type\",...}]"} (a JSON string nested inside JSON) instead of a real array. src/app/api/students/route.ts already returns { error: e.issues } for the same case, so this introduces a shape inconsistency across the API. Not a functional bug, but worth returning error.issues (when it's a ZodError) instead of .message so all validation-error responses have the same shape. Non-blocking — fine to fix as a follow-up.

david-fs-valente (approved):

src/services/apiResponse.ts
dinisjcorreia: Addressed in 0ed2c63. `errorResponse` now detects `ZodError` and returns `error.issues` as a real JSON array, matching the existing students route. Other `Error` instances still return `.message`; arbitrary values remain safely stringified. Runtime checks cover all three cases.

PR #165, opened 2026-07-12
fix: fix html lang and run accessibility pass

### LINKED ISSUE #141: [Fix] Fix html lang and run an accessibility pass
### Description
The root layout sets `<html lang="en">` on a Portuguese app, and wider accessibility is unaudited.

### Why
Screen readers and translation tools mis-handle all content when the language is wrong.

### How
Fix to `lang="pt"` now; then run an a11y pass (focus states, form labels, colour contrast, keyboard navigation).

### Dependencies
None (the lang fix precedes the i18n issue).

david-fs-valente:
## Summary
- Set `<html lang>` to `pt` (was `en`) to match the site's Portuguese content
- Accessibility pass: keyboard navigation (unfocusable click handlers, misplaced handlers), form label/id association, dialog and accordion ARIA, icon-only button labels, skip-to-content link

Closes #141

### DIFF
--- a/src/components/Input/index.tsx
+++ b/src/components/Input/index.tsx
@@ -20,25 +20,28 @@ const Input: React.FC<InputProps> = ({
   disabled,
   inputRef,
   type = "text",
+  id,
   ...rest
 }) => {
-  const id = slugifyId(name);
+  const resolvedId = id || slugifyId(name) || undefined;
 
   return (
     <div className="flex w-full flex-col">
-      <label
-        className={`mb-1 text-sm font-normal text-white ${
-          center ? "text-left" : ""
-        }`}
-        htmlFor={id}
-      >
-        {name}
-      </label>
+      {name && (
+        <label
+          className={`mb-1 text-sm font-normal text-white ${
+            center ? "text-left" : ""
+          }`}
+          htmlFor={resolvedId}
+        >
+          {name}
+        </label>
+      )}
       <input
         type={type}
         name={name}
         disabled={disabled}
-        id={id}
+        id={resolvedId}
         placeholder={placeholder}
         ref={inputRef}
         className={`h-14 w-full border border-white/35 bg-[#141414] px-2 py-1 text-sm text-white placeholder:text-white/50 focus:border-primary focus:ring-0 disabled:text-gray-600 ${className}`}
--- a/src/components/Input/index.tsx
+++ b/src/components/Input/index.tsx
@@ -1,5 +1,7 @@
 "use client";
 
+import slugifyId from "@/utils/slugifyId";
+
 interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {
   name: string;
   center?: boolean;
@@ -20,13 +22,15 @@ const Input: React.FC<InputProps> = ({
   type = "text",
   ...rest
 }) => {
+  const id = slugifyId(name);
+
   return (
     <div className="flex w-full flex-col">
       <label
         className={`mb-1 text-sm font-normal text-white ${
           center ? "text-left" : ""
         }`}
-        htmlFor={name}
+        htmlFor={id}
       >
         {name}
       </label>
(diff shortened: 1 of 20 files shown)


src/components/Modal/index.tsx
```diff
@@ -24,8 +45,12 @@ const Modal: React.FC<ModalProps> = ({
         onClick={() => setIsVisible(false)}
       ></div>
       <main
+        role="dialog"
+        aria-modal="true"
```
dinisjcorreia, line 78: [P2] `aria-modal="true"` promises modal interaction, but focus is only moved to the container. Shift+Tab can immediately reach controls behind the overlay, and repeated Tab eventually leaves the dialog. Use native `<dialog>.showModal()` or trap Tab and make background content inert.

src/components/TopBar/index.tsx
```diff
@@ -48,6 +48,7 @@ const TopBar: React.FC = () => {
           {!session.user ? (
             <Link
               href={isAuthPage ? "/" : "/login"}
+              aria-label="Iniciar sessão"
```
dinisjcorreia: [P2] On `/login`, `/signup`, and `/password-reset`, this link points to `/`, but its accessible name remains “Iniciar sessão”. Screen-reader users receive the wrong destination. Set a conditional home-page label or render a distinct home link.

dinisjcorreia (changes requested):
Accessibility review: requesting changes.

Two additional blocking findings are attached inline:
- modal focus can escape into obscured page content;
- auth-page home link is announced as login.

Also fix the duplicate generated field IDs and unnamed dialog already identified by Copilot. Please review and resolve all Copilot comments before requesting re-review.

Verification: production build and `git diff --check` pass. Repository lint config currently crashes; standalone TypeScript check reports unrelated existing errors; Prettier reports changed-file formatting issues.

src/components/Modal/index.tsx
```diff
@@ -1,20 +1,70 @@
...
+import React, { Dispatch, SetStateAction, useEffect, useRef } from "react";
 
 interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
   isVisible: boolean;
   setIsVisible: Dispatch<SetStateAction<boolean>>;
   children: React.ReactNode;
 }
 
+const FOCUSABLE_SELECTOR =
+  'a[href], button:not([disabled]), textarea:not([disabled]), input:not([disabled]), select:not([disabled]), [tabindex]:not([tabindex="-1"])';
```
dinisjcorreia: [P2] This selector includes hidden inputs. In the avatar modal, `AvatarCropper` renders `<input hidden>` before the Confirmar button. It becomes `first`, so Tab from Confirmar calls `first.focus()` on an unfocusable node, while Shift+Tab from Confirmar misses the wrap condition and can leave the dialog. Exclude hidden/non-rendered elements (or use native `<dialog>.showModal()`).

src/components/Input/index.tsx
```diff
@@ -18,27 +20,32 @@ const Input: React.FC<InputProps> = ({
...
     <div className="w-full flex flex-col">
-      <label
-        className={`text-sm font-normal text-white mb-1 ${
-          center ? "text-left" : ""
-        }`}
-        htmlFor={name}
-      >
-        {name}
-      </label>
+      {name && (
```
dinisjcorreia, line 30: [P2] Omitting the generated label/ID avoids duplicates, but every `<Input name="">` in `ProfileSection` is now an unnamed control: the visible outer labels (Nome, Ano, Email, LinkedIn, Github) have no `htmlFor`. Give each caller a unique `id` and connect its outer label, or pass a real accessible name.

dinisjcorreia (changes requested):
Re-review completed. The previous dialog naming, duplicate-ID, QR button type, className, and TopBar label findings are addressed.

Two accessibility blockers remain and are attached inline:
- focus trap treats hidden controls as focusable, so keyboard focus can still escape;
- profile fields with `name=""` now avoid duplicate IDs but remain unnamed because their visible outer labels are not associated.

Verification: production build and `git diff --check` pass. Repository lint config still crashes, standalone TypeScript still reports existing project errors, and Prettier still reports changed-file formatting issues. Please also ensure all Copilot threads are resolved before requesting another review.

dinisjcorreia (approved):
Re-review complete. Both remaining accessibility blockers are resolved: hidden controls are excluded from the modal focus trap, and profile labels are associated with unique input IDs.

Verified: CI Typecheck & Lint passes; local production build, lint, Prettier, and `git diff --check` pass. Local standalone typecheck only fails because Git-LFS image assets are absent from this checkout.

Approved.

Issue #151, opened 2026-07-12
[Fix] Fix SavedStudent uniqueness grain (prevent duplicate company saves)


david-fs-valente:
### Description
A student can be saved multiple times by the same company, and the app-level dedup check races with no DB backstop.

### Why
The primary key is `(studentId, employeeId)` (employee-scoped) but the business rule dedups per company (`isSaved` checks companyId).

### How
Enforce a company-scoped unique constraint (add `companyId` + `@@unique([studentId, companyId])`) or change the dedup to employee grain. Ship as a migration.

### Dependencies
Depends on #138 (Prisma migrations) — this is a schema change and must ship as a reviewable migration.

PR #167, opened 2026-07-12
fix: wrap dependent route writes in transactions

### LINKED ISSUE #137: [Fix] Wrap multi-write route handlers in transactions
### Description
Route handlers perform several dependent writes with no atomicity, so a partial failure leaves inconsistent data.

### Why
`prisma.$transaction` is used 0 times; `students` POST does 3 writes, and signup/employee, user, saved, companies do 2 each.

### How
Wrap dependent write sequences in `prisma.$transaction`; the service method owns the transaction boundary (lands with the service-layer refactor).

### Dependencies
None (can wrap existing handlers now).

dinisjcorreia:
## Summary
- wrap student, employee signup, company, saved-student, and company-interest write sequences in Prisma transactions
- make action completion idempotent with an atomic compound-key upsert
- return explicit JSON errors for transaction failures in saved-student and student update routes
- include sibling student profile and CV update flows with the same atomicity issue

## Verification
- `pnpm test` (6 passing)
- `git diff --check`
- touched files pass targeted TypeScript checking; repository has pre-existing type errors elsewhere
- dev-server probes confirm explicit 401 JSON responses for all three reviewed routes
- authenticated transaction paths require a configured Supabase session and database
- lint unavailable because existing ESLint config crashes with `Converting circular structure to JSON`


david-fs-valente (changes requested):
Reviewed the transaction-wrapping changes. The atomicity fix itself is correctly wired (every write inside each `$transaction` block consistently uses `tx`, no accidental calls to the outer `prisma` client), but two related issues combine into a real risk and should be addressed before merge.

**1. A benign race in `completeAction` can now fail the primary write, not just the side effect.**

`completeAction` ([src/lib/completeAction.ts:42-51](https://github.com/Nucleo-Estudantes-Informatica-ISEP/fallstack-website/blob/683c35f/src/lib/completeAction.ts#L42-L51)) does a `findFirst` already-completed check, then `create`s the `ActionCompletion` row — inside the caller's transaction now. `ActionCompletion` has `@@unique([actionId, studentId])`, so if two concurrent requests race on the same student+action, the loser's `create` throws `P2002`. Because that call now runs inside the same transaction as the CV upload / saved-student insert / profile update, the *entire* write rolls back with it.

Before this PR, the primary write had already committed by the time `completeAction` ran, so this race only ever broke the gamification side effect — never the user-facing action. Now a transient dedup race in the action-completion system can fail a CV upload, a student save, or a profile edit. This is a reasonable trade (atomicity over partial writes) but it's a real behavior change and, combined with point 2 below, means that failure surfaces as a raw unhandled error rather than a clean response.

**2. No `try/catch` around the transaction in three of the touched routes.**

- [src/app/api/saved/route.ts:112](https://github.com/Nucleo-Estudantes-Informatica-ISEP/fallstack-website/blob/683c35f/src/app/api/saved/route.ts#L112) (`POST`)
- [src/app/api/students/[code]/cv/route.ts:97](https://github.com/Nucleo-Estudantes-Informatica-ISEP/fallstack-website/blob/683c35f/src/app/api/students/%5Bcode%5D/cv/route.ts#L97) (`POST`)
- [src/app/api/students/[code]/route.ts:66](https://github.com/Nucleo-Estudantes-Informatica-ISEP/fallstack-website/blob/683c35f/src/app/api/students/%5Bcode%5D/route.ts#L66) (`PATCH`)

If the transaction throws (e.g. the P2002 race above), the exception is unhandled and the client gets Next.js's generic error response instead of a JSON `{ error }` body — inconsistent with AGENTS.md's stated convention of `{ error: string }` with an explicit status code on every failure. This isn't a new gap (these routes had no try/catch before either), but it's meaningfully more likely to trigger now that a secondary system's race can trip the primary write.

**Requested change:** wrap these three handlers' transaction calls in try/catch and return a proper `{ error }` JSON response (status 500, optionally special-casing Prisma `P2002` if you want a friendlier message for the dedup race).

david-fs-valente:
**Suggestion (non-blocking):** `SavedStudent`'s known grain mismatch — the DB PK is employee-scoped (`[studentId, employeeId]`), while `isSaved()` in `src/lib/savedStudents.ts` dedupes at the company level (`savedBy: { companyId }`), per AGENTS.md — is untouched by this PR. The new transaction doesn't fix or worsen it (two employees at the same company can still both save the same student), just flagging it as still-open follow-up work, not something this PR needs to address.

dinisjcorreia:
Addressed in 1041c61:

- Replaced the `findFirst` + `create` sequence in `completeAction` with an atomic compound-key `upsert`, removing the P2002 dedup race.
- Added explicit transaction error handling to saved-student POST, student CV POST, and student profile PATCH.
- Saved-student P2002 now returns `{ error: "Student already saved" }` with HTTP 409; other failures return explicit JSON errors with HTTP 500 and are logged.
- Added a focused `completeAction` upsert test; full test suite passes (6/6).

The SavedStudent employee/company grain mismatch remains unchanged, as noted as non-blocking/out of scope.

src/app/api/user/route.ts
```diff
@@ -26,7 +26,7 @@ export async function PATCH(req: NextRequest) {
     });
 
     // Update interests for all employees in the company
-    await Promise.all(
+    await prisma.$transaction(
```
david-fs-valente: [P2] This transaction has no error handling. Every other route touched in this PR wraps its `$transaction` in try/catch and returns an explicit JSON error on failure (see `saved/route.ts`, `students/[code]/route.ts`, `students/[code]/cv/route.ts`). Issue #137 explicitly calls out this route ("signup/employee, user, saved, companies do 2 each"), so it's in scope here. As written, if this transaction throws, the client gets an unhandled exception / non-JSON error response instead of the consistent JSON contract established elsewhere in this PR. Please wrap in try/catch + `reportError`, matching the other routes.

src/app/api/auth/signup/employee/route.ts
```diff
@@ -41,27 +41,26 @@ export async function POST(req: Request) {
 
     const supabaseUser = data.user;
 
-    // Create application user with EMPLOYEE role if not existing
-    try {
-      await prisma.user.create({
-        data: {
+    await prisma.$transaction(async (tx) => {
```
david-fs-valente: [P2] The Supabase auth user is created via `supabase.auth.signUp` (line 28) before this transaction opens. If `tx.employee.create` fails (e.g. duplicate id), the Supabase auth account isn't rolled back and is left orphaned with no corresponding `User`/`Employee` row. Not introduced by this diff, but since this PR's purpose is closing atomicity gaps in this exact flow, worth addressing here too — e.g. delete the Supabase auth user on transaction failure.

david-fs-valente (changes requested):
Reviewed the transaction-wrapping changes across all six touched routes plus the `completeAction` idempotency fix. Verified against PR head (43aeb20) in a disposable worktree: typecheck clean, `pnpm test` 6/6 passing.

The `completeAction` atomic-upsert fix (keyed on the real `actionId_studentId` compound constraint) is solid and backed by a unit test. Most routes now correctly wrap their dependent writes and return a consistent JSON error on transaction failure.

Requesting changes on two gaps that fall within this PR's own scope (issue #137 explicitly lists `user` as one of the routes to wrap) — both left as inline comments:

- `/api/user` PATCH has no error handling around its new `$transaction`, inconsistent with every other route this PR touches.
- Employee signup: the Supabase auth user created via `signUp` isn't covered by the new transaction boundary, so a Prisma failure orphans the auth account.

Note: the SavedStudent company-scoped uniqueness race (two different employees at the same company concurrently saving the same student) is real, but out of scope here since it's tracked separately in #151 — not blocking this PR.

dinisjcorreia:
Addressed both review threads in 55efc6a:

- `/api/user` now catches transaction failures, logs via `reportError`, and returns explicit JSON with HTTP 500.
- Employee signup now compensates a failed Prisma transaction by deleting the newly created Supabase auth user through the service-role admin client. Cleanup failures and transaction failures are logged separately; client receives explicit JSON with HTTP 500.

Verification: `pnpm exec tsc --noEmit` clean, `pnpm test` 6/6, touched-file ESLint clean, `git diff --check` clean. Repo-wide lint still has pre-existing/generated-file failures.

david-fs-valente (approved):
Re-reviewed after the follow-up commits — both findings from my previous review are resolved, and I re-verified rather than trusting the commit messages:

- **`completeAction` race** — fixed properly: `actionCompletion.upsert()` on the compound unique key `actionId_studentId` replaces `findFirst`+`create`, closing the TOCTOU window at the DB level instead of just catching the resulting error. Covered by a new unit test.
- **Missing try/catch** — added to all the routes I flagged (`saved/route.ts`, `students/[code]/cv/route.ts`, `students/[code]/route.ts`, plus `user/route.ts`), each returning proper `{error}` JSON with correct status codes; `saved/route.ts` correctly special-cases Prisma `P2002` → 409 for the residual save-conflict race.
- Bonus: employee signup now best-effort rolls back the Supabase auth user if the DB transaction fails, avoiding orphaned auth accounts.

Independently verified: `pnpm test` 6/6 pass, `tsc --noEmit` clean on every touched file (repo-wide remaining errors are pre-existing asset-resolution issues, unrelated to this diff), targeted eslint on all 8 touched files clean.

Approving.

david-fs-valente:
**Notable side-fix worth flagging explicitly:** the "align company transaction with schema" commit removed `user: { connect: { id: userId } } }` from the `Company.create()` call in `src/app/api/companies/route.ts`. I initially assumed this was cosmetic (it fixed a pre-existing TS error), but I verified it empirically against a real Postgres instance and it's not cosmetic — it's a live bug fix.

`Company` has no `user` relation field in the schema at all (confirmed via the generated `CompanyCreateInput`/`CompanyUncheckedCreateInput` types, which don't include `user`). Running the exact pre-PR `prisma.company.create()` call against a throwaway DB with the current schema throws `PrismaClientValidationError` immediately:

```
Invalid `prisma.company.create()` invocation:
...
    user: {
    ~~~~
      connect: {
```

That means **on `dev`, `POST /api/companies` currently fails on every request** — company signup is completely broken. This PR fixes it as an incidental side effect of the schema-alignment commit, separate from the transaction-atomicity work it set out to do. Worth calling out in the changelog/release notes as a fix in its own right, since it's arguably higher-impact than the atomicity work — and worth a quick manual signup smoke test post-merge to confirm end-to-end.

PR #168, opened 2026-07-12
refactor: introduce service and repository layers

### LINKED ISSUE #126: [Refactor] Introduce a service/repository layer with thin route handlers
### Description
There is no clear layering: route handlers are fat, Prisma is called directly from routes/pages/lib, and server and client code are mixed in the same folders.

### Why
E.g. `students/route.ts` POST is ~120 lines (auth, validation, creation, storage), and save logic is duplicated across three routes. This makes the code hard to test and to extract.

### How
Add `application/repositories/` (the only place touching Prisma) and `application/services/` (called by routes and Server Components). Routes become auth -> parse -> service -> respond. Move client fetch wrappers to a client-only folder; enforce the boundary with `import "server-only"`.

### Dependencies
None. This is a hub that unblocks the DTO, testing, and HTTP-layer issues.

### LINKED ISSUE #133: [Refactor] Clean up data-layer smells (naming, return shapes, magic strings)
### Description
The `lib/` data layer has inconsistent naming, leaky return shapes, magic strings, and dead code.

### Why
`fetch*` vs `get*` naming with file/export mismatches; `fetchStats` returns `[savedCount, savedCount]`; the ISEP email regex and domain are duplicated; `info@nei-isep.org` is hardcoded; `exclude()` is unused; signup has swallowed catches.

### How
Standardize naming; return named objects; centralize the ISEP email validator and domain; delete dead code; handle swallowed errors. Most of this lands with the service-layer refactor.

### Dependencies
None (co-lands with the service-layer and Value-Object work).

dinisjcorreia:
## Summary
- isolate all runtime Prisma access in application repositories protected by `server-only`
- move route and Server Component business logic into application services, including one shared student-save flow
- move browser fetch wrappers into a `client-only` API folder
- enforce boundaries through structural tests and ESLint restricted imports

## Review follow-up
- admin saves now delegate to `saveStudent()`
- profile and CV access share one authorization policy
- server boundary coverage includes `src/lib/http/`
- company matching query now lives in `companyRepository` with inferred types and no `any`
- focused tests cover save deduplication, booth lookup, student access, and company ranking
- HTTP error response helper moved to the documented `src/lib/http/server.ts` location

## Verification
- `pnpm test` — 11/11 passed
- `pnpm build` — passed
- `pnpm exec eslint src/application src/lib/http` — passed
- `git diff --check` — passed

## Notes
- signup requests containing legacy `role: "COMPANY"` previously attempted an invalid Prisma enum write that was swallowed; they now persist the valid `EMPLOYEE` role. Current frontend callers do not send this field.
- standalone `tsc` still reports pre-existing project errors; no new errors originate from the refactored modules.


david-fs-valente (changes requested):
## Review: service/repository layering refactor

Overall this is a solid, low-regression-risk refactor — I verified independently (clean `pnpm install`, `pnpm build`, `pnpm test`, and `pnpm exec tsc --noEmit` actually shows *fewer* type errors than `dev`, not more). The repository naming convention is a genuine hygiene win and the booth-`switch` → data-map change is a nice OCP improvement. Requesting changes on a few specific gaps before merge, mainly around SOLID/DRY in the new service layer and the boundary tests the PR itself introduces.

### Blocking

1. **Save-flow consolidation is incomplete.** `saveStudent()` correctly unifies the scan-save (`POST /api/saved`) and manual-save (`PATCH /api/saved`) paths. But `saveStudentAsAdmin()` (`/api/admin/save-student`) reimplements the same "already saved by this company" check independently — `findSavedStudent(studentId, companyId)` vs. `isStudentSaved(companyId, code)` in `saveStudent()`. Same invariant, two implementations that can drift. Please resolve the student in `saveStudentAsAdmin` and delegate into `saveStudent()` instead of duplicating the check.

2. **Duplicated authorization predicate for student PII/CV access.** `studentService.getStudentProfile()` and `studentService.getStudentCv()` both hand-roll the identical access rule:
   ```ts
   access.studentCode === code || access.isAdmin ||
     (!!access.companyId && (await isStudentSaved(access.companyId, code)))
   ```
   This guards PII exposure — please extract a single `isAllowedToViewStudent(code, access)` helper and use it in both places so the rule can't silently diverge between the two call sites.

3. **Boundary test doesn't cover the folder this PR introduces.** `src/application/boundaries.test.ts`'s "server and client modules declare their boundary" test only walks `application/{repositories,services}` and `client/`. This PR also adds `src/server/http/httpErrorResponse.ts` with its own `import "server-only"`, but that folder isn't checked. Given the PR description explicitly says the tests should catch "missing server/client markers," please extend the test to cover `src/server/**` too (or fold `httpErrorResponse.ts` into a directory the test already watches).

### Non-blocking, please consider

4. **Misplaced repository responsibility:** `findInterestMatchingCompanies()` lives in `interestRepository.ts` but only queries `prisma.company`/`prisma.user`, never `prisma.interest`. Move it to `companyRepository.ts`, which already owns company-fetching queries.

5. **New `as any` cast:** `companyService.getInterestMatchingCompanies()` casts `(companies as any[])` to work around `findInterestMatchingCompanies`'s untyped merged return shape. Worth typing the return properly instead of casting — this adds debt in a codebase already trying to reduce `as any` usage.

6. **Prisma-leak test is regex-over-text**, so it's easy to defeat by aliasing the import (`import db from "../repositories/database"`) and doesn't fire outside `pnpm test`. An ESLint `no-restricted-imports` rule scoped to `@/application/repositories/database` would enforce this at edit time for every contributor rather than relying on a naming convention being followed by hand.

7. **No unit tests for the newly extracted service logic.** The two new tests are structural only (no Prisma outside repositories / markers present). None of `saveStudent`'s dedup+booth-completion logic, `getStudentProfile`'s access predicate, or `getInterestMatchingCompanies`'s ranking is covered, even though extracting them into standalone functions is what makes them testable for the first time.

8. **Undocumented behavior change riding along with the refactor:** `src/app/api/auth/signup/route.ts` used to write `role: "COMPANY"` (an invalid enum value that a swallowed `catch` silently discarded) and now writes `"EMPLOYEE"`. I confirmed no current frontend caller actually sends `role` on this endpoint today, so it's low risk, but it is a real semantic change and isn't called out in the PR description — worth a one-line mention so it isn't rediscovered later by diffing.

9. Small placement nit: `httpErrorResponse.ts` sits at `src/server/http/`, while the documented target architecture designates `src/lib/http/server.ts` for this kind of concern. Not urgent, but whoever builds out the real HTTP layer next will find this module already parked somewhere else.

dinisjcorreia:
@david-fs-valente Addressed all review points in `abcc37e`:

- unified admin/manual/scan save invariant through `saveStudent()`
- extracted one student PII/CV authorization policy
- expanded server boundary coverage and added ESLint restricted-import enforcement
- moved company matching query to `companyRepository`; removed `as any`
- added focused logic tests for deduplication, booth selection, access, and ranking
- documented the signup role behavior change
- moved HTTP helper to `src/lib/http/server.ts`

Verification: 11/11 tests, scoped ESLint, production build, and diff checks pass. This review contains no inline review threads, so GitHub exposes no resolvable conversations; requesting re-review.

dinisjcorreia:
Ported the non-superseded data-contract work from #174 in commit d3c30d5:

- named stats contract
- centralized ISEP email validation/normalization + tests
- idempotent user upserts
- signup error propagation
- centralized NEI contact email

Verification: 13 tests pass; production build passes.

david-fs-valente (approved):

david-fs-valente (commented):
Reviewed the service/repository refactor. This is a clean, well-executed structural change — no blocking issues, just findings worth being aware of.

**Verified claims:**
- `pnpm test`: 13/13 pass (description says 11/11 — stale, harmless). Tests are substantive (save dedup, booth lookup, access policy, company ranking), not superficial.
- `pnpm build` and `pnpm exec eslint src/application src/lib/http`: both pass clean.
- `pnpm exec tsc --noEmit`: zero errors on this branch *and* current `dev` — the PR description's "pre-existing errors remain" note is now stale (dev was cleaned up by other merged work since this PR was opened), not a defect introduced here.

**Security note on `role: "COMPANY"` — non-issue, description overstates it.** Checked before/after: on `dev`, `signup/route.ts` already did `.omit({ role: true })`, so client-supplied `role` never reached Prisma — the write was always hardcoded `role: "STUDENT"`. Here, `authApplicationService.ts:13-16,39-51` types `role` as `"STUDENT" | "EMPLOYEE"` and both call sites pass hardcoded literals. No client input ever reaches the `role` column, before or after this PR.

**Non-blocking finding — coordinate with #167:** `grep -rn "\$transaction" src/` returns nothing on this branch — zero transactions. `savedStudentService.ts:60-76` (`saveStudent()`) and `studentService.ts:37-63,89-95` (`createStudentProfile`, `setStudentCv`) do sequential, un-transacted writes, which is exactly the atomicity gap #167 fixes — but for the old file locations. Since this PR moves that logic into new service files, #167's diff won't apply cleanly here regardless of merge order (the file moves are too extensive for a textual merge to relocate the fix correctly). Suggested path: merge this PR first, close #167 without merging, and reapply the transaction-wrapping fix as a small follow-up PR directly against `savedStudentService.ts` / `studentService.ts` / `authApplicationService.ts`.

**Improvement over what #167 flagged:** the migrated routes now consistently wrap calls in try/catch → `httpErrorResponse()` (`src/lib/http/server.ts`), returning proper `{error}` JSON with correct status via an `HttpError` class — this resolves the "no try/catch, generic error page" gap raised in the #167 review. Nice.

**Boundary enforcement is solid:** all `repositories/*.ts` files import `server-only`; the `no-restricted-imports` ESLint rule (`eslint.config.mjs:67-92`) is correctly scoped to ban `PrismaClient`/raw-prisma imports everywhere except `src/application/repositories/**`, and no runtime leaks were found outside that boundary.

Sampled `saved/route.ts`, `students/[code]/cv/route.ts`, and `admin/save-student/route.ts` against their pre-PR versions — auth checks, status codes, and validation are all preserved; admin save-student now correctly delegates to `saveStudent()` as claimed.

Not requesting changes — the missing transactions are a merge-order/follow-up item, not a regression this PR introduces.

PR #169, opened 2026-07-12
fix: docker health check

### LINKED ISSUE #144: [Fix] Add /api/health route for the Docker healthcheck
### Description
The container is perpetually reported "unhealthy" because the healthcheck hits a route that does not exist.

### Why
The Dockerfile `HEALTHCHECK` curls `/api/health`, which is missing, so the orchestrator may fail deploys or restart-loop.

### How
Add a trivial `GET /api/health` returning 200 (or repoint the healthcheck). Verify against the running container.

### Dependencies
None.

Pedro-Sousa7:
- Changed dockerfile to fix build errors
- Added /api/health which returns `Status 200` for the docker


next.config.js
```diff
@@ -27,6 +27,7 @@ const nextConfig = {
   eslint: {
     ignoreDuringBuilds: true,
   },
+  output: "standalone",
```
david-fs-valente: Duplicate key: `output: "standalone"` is already set at the top of `nextConfig` on `dev` (line 23). This adds it a second time in the same object literal — dead code / `no-dupe-keys` lint violation. Safe to drop this line.

david-fs-valente (approved):
Health check fix looks good and unblocks the Docker HEALTHCHECK. One small cleanup noted inline.

david-fs-valente (approved):

PR #172, opened 2026-07-12
refactor(auth): consolidate password reset on Supabase

### LINKED ISSUE #132: [Refactor] Consolidate password reset onto a single Supabase flow
### Description
Password-reset logic was implemented multiple times; standardize on Supabase Auth.

### Why
Supabase reset, a custom `passwordResetToken` flow, and an admin password-change all coexisted — confusing and a larger attack surface. (The leaking custom route is already removed.)

### How
Standardize on the Supabase flow; drop the now-unused `passwordResetToken` model (as a migration). Document that all auth goes through Supabase Auth.

### Dependencies
Depends on the account-takeover and reset-token security fixes, which are already merged.

dinisjcorreia:
## Summary

- keep password recovery on one Supabase Auth flow: reset email, recovery session verification, then `auth.updateUser`
- remove the admin password override route, page, component, and validation schema
- add an idempotent migration dropping the legacy `PasswordResetToken` table
- document Supabase Auth ownership and add an architecture regression test

## Testing

- `pnpm test` (6 tests pass)
- `git diff --check`
- `pnpm exec tsc --noEmit` (blocked by existing unrelated repository type errors; changed files report no errors)
- targeted ESLint (blocked by existing circular ESLint config error)

## Database

Migration was not executed locally. Apply `prisma/migrations/20260712190000_drop_password_reset_token/migration.sql` through the project database workflow.

Closes #132


src/app/api/auth/password-reset/route.ts
```diff
@@ -9,78 +9,34 @@ const requestResetSchema = z.object({
...
-      );
+      return NextResponse.json({ error: "Invalid email" }, { status: 400 });
     }
 
-    const { error } = await supabase.auth.updateUser({ password });
+    const redirectTo = new URL(
+      "/password-reset/confirm",
+      req.nextUrl.origin
+    ).toString();
+    const { error } = await supabase.auth.resetPasswordForEmail(
```
david-fs-valente, line 22: [Blocking] This client is built with plain `createClient` from `@supabase/supabase-js` (not `@/utils/supabase/server`), so `flowType` defaults to `'implicit'` (auth-js `GoTrueClient`'s default). `resetPasswordForEmail`'s own implementation only generates and sends a PKCE `code_challenge` when `this.flowType === 'pkce'` — so as written, this call makes GoTrue mint an **implicit-flow** recovery link (`#access_token=...&type=recovery` hash fragment), not a PKCE `?code=` link.

The confirm page uses `@/utils/supabase/client`'s `createClient()`, which wraps `@supabase/ssr`'s `createBrowserClient` — that defaults `flowType: 'pkce'`. When that page's `detectSessionInUrl` sees an implicit-style URL against a pkce-configured client, `GoTrueClient._getSessionFromURL` throws `AuthPKCEGrantCodeExchangeError('Not a valid PKCE flow url.')` (explicit mismatch guard in auth-js), no session gets set, and the follow-up `updateUser({password})` fails with "Auth session missing." Password reset won't work for any user with this combination.

Suggested fix: build this client from `@/utils/supabase/server` (already used elsewhere, e.g. employee signup) instead of a raw `@supabase/supabase-js` client, so it shares the same PKCE flow type as the browser client on the confirm page.

src/app/(auth)/password-reset/confirm/page.tsx
```diff
@@ -67,50 +51,11 @@ const PasswordResetForm: React.FC = () => {
...
-        });
-
-        if (!res.ok) {
-          setLoading(false);
-          const json = await res.json();
-          return setPwError(json.error || "Algo correu mal");
-        }
-      }
+      const supabase = createClient();
+      const { error } = await supabase.auth.updateUser({ password });
```
david-fs-valente, line 56: [Blocking] Other half of the flow-type mismatch flagged on `password-reset/route.ts`: this client defaults to `flowType: 'pkce'`, but the recovery link generated by the request route is implicit-flow. `detectSessionInUrl` will fail to establish a session before this line runs, so `updateUser` will throw "Auth session missing" here for every real recovery attempt. Worth noting the new `authFlow.test.ts` only asserts source-text patterns, so it doesn't catch this — the regression test doesn't cover the actual runtime behavior it's meant to protect.

david-fs-valente (changes requested):
Verified against PR head (e3233c6) in a disposable worktree: typecheck clean, `pnpm test` 6/6 passing. Checked open issues — this PR closes #132, and I couldn't find any issue tracking the specific defect below, so it's blocking here rather than deferred.

Requesting changes: the request-side and confirm-side Supabase clients use different auth flow types, which will break password recovery end-to-end. Details and a suggested fix are inline on both ends of the flow.

david-fs-valente:
Minor, non-blocking (leaving as a plain comment rather than part of the review): once the flow-type mismatch flagged in the review is fixed, it'd be worth a short note — in code or in AGENTS.md — on why `@/utils/supabase/client` and any future server-side Supabase Auth callers must consistently use the PKCE flow. This class of bug is easy to reintroduce silently: both flow types compile and typecheck fine, and nothing short of actually clicking a real recovery email surfaces the mismatch.

src/app/api/auth/password-reset/route.ts
dinisjcorreia: Fixed in 0ad0ec0: reset requests now use the shared SSR Supabase client, so request and confirm paths both use PKCE with cookie-backed verifier storage. Added a regression check for the shared client and runtime flow-type equality.

src/app/(auth)/password-reset/confirm/page.tsx
dinisjcorreia: Fixed in 0ad0ec0. The request side now uses the shared PKCE SSR client, and the regression test now verifies SSR/browser flow types match at runtime.

david-fs-valente (approved):
Re-reviewed after the flow-type mismatch fix in 0ad0ec0 — verified it independently rather than taking the fix on faith.

**The fix is architecturally sound, not just superficially matching.** I traced this past what's visible in the diff: the PKCE `code_verifier` cookie set during the server-side `resetPasswordForEmail()` call is stored via `@supabase/ssr`'s default cookie options (`httpOnly: false`, `sameSite: "lax"`, `path: "/"` — confirmed by reading the installed package's `constants.js`), so it's genuinely readable by the browser client on the confirm page after the redirect. Had it been `httpOnly`, the fix would look correct on paper but still fail silently in production. It isn't — the fix holds up end to end, not just at the `flowType` level the regression test checks.

Also verified:
- `pnpm test`: 6/6 pass. `pnpm exec tsc --noEmit`: zero errors in any touched file (remaining repo-wide errors are pre-existing, unrelated asset-resolution issues).
- No dangling references anywhere in the tree to the removed admin change-password route/page/component.
- The migration is genuinely a no-op for fresh environments — `20260712000000_init` never creates `PasswordResetToken` in the first place, so `DROP TABLE IF EXISTS` is purely defensive for legacy databases, consistent with the "idempotent" claim.

Approving.

david-fs-valente:
**Non-blocking aside, pre-existing and out of scope for this PR:** `src/utils/supabase/server.ts` (not touched here) builds its shared "SSR client for auth/session" using `serverEnv.SUPABASE_SERVICE_ROLE_KEY` — despite the code comment claiming "public anon key." This PR routes the previously anon-key-scoped, fully unauthenticated `POST /api/auth/password-reset` request through that client for the first time (replacing an explicit anon-key `@supabase/supabase-js` client).

It doesn't break anything functionally — the `/recover` GoTrue endpoint behaves the same regardless of which key authenticates the call, and this route never calls `admin.*` — but it does mean an anonymous, public endpoint now runs on the same client instance that carries the most privileged credential in the project. This is already the pattern for login/signup/logout too, so it's systemic rather than introduced here. Worth tightening (fix the comment or actually scope this wrapper to the anon key) as a follow-up, independent of this PR.

PR #173, opened 2026-07-12
refactor: extract edition configuration layer

### LINKED ISSUE #129: [Refactor] Extract an edition/ config layer (single-app edition model)
### Description
Per-edition data is scattered and partly hardcoded in logic, so a new edition still requires a code fork.

### Why
The booth switch, `actionNames`, and the sponsor/company/schedule/FAQ utils hold this year's data in code.

### How
Consolidate into an `edition/` config module (sponsors, tiers, actions, schedule, FAQ, branding) so a new edition is a config change. This also enables replacing the booth switch with a lookup.

### Dependencies
None. Unblocks the booth-lookup and sponsors issues.

dinisjcorreia:
## Summary
- consolidate sponsors, company tiers, schedule, FAQ, actions, and branding under `src/edition`
- replace booth action switch with normalized edition lookup
- drive metadata, manifest, dates, storage keys, exports, and pass branding from edition config
- add focused booth lookup coverage

## Verification
- `pnpm test` (6 passed)
- `pnpm build`
- `git diff --check`

## Known repository tooling issues
- `pnpm exec eslint ...` fails before linting because the existing ESLint 9 config throws `TypeError: Converting circular structure to JSON`
- standalone `tsc --noEmit` reports existing project-wide type errors; production build is configured to skip type validation


src/edition/actions.ts
```diff
@@ -0,0 +1,70 @@
...
+  glinttBooth: "Banca Glintt",
+  konkConsultingBooth: "Banca konk consulting",
+};
+
+const boothActions: Record<string, string> = {
+  akapeople: actionNames.akaPeopleBooth,
+  natixis: actionNames.natixisBooth,
+  apr: actionNames.aprBooth,
+  "apr - technology solutions": actionNames.aprBooth,
+  "hitachi solutions": actionNames.hitachiBooth,
```
david-fs-valente, line 26: [Blocking] `DiamondCompanies.tsx` sets this company's display/registration-facing name to `"hitachi"` (no "solutions" suffix), but this map only has a key for `"hitachi solutions"`. Confirmed directly:

```
getBoothActionName("hitachi")            -> undefined
getBoothActionName("hitachi solutions")  -> "Banca Hitachi"
```

If Hitachi's `Company.name` in the DB matches their own display config (plausible — same as how the APR and msg insur:it mismatches just above were fixed), scanning a student at their booth silently completes no action: no error, no log, just a no-op `if (boothActionName)` skip in `/api/saved`. Given this PR already fixed the identical pattern for two other companies, this looks like a missed instance rather than an intentional exclusion. Suggest adding `"hitachi"` as an alias here, and a matching case in `actions.test.ts`.

david-fs-valente (changes requested):
Verified against PR head (044091a) in a disposable worktree: typecheck clean, `pnpm test` 6/6 passing. Checked open issues — this PR closes #129, and I couldn't find any issue tracking the specific defect below, so it's blocking here.

Good refactor overall: the switch→lookup rewrite is a real correctness fix (case-insensitive matching), and it already fixes two analogous name-mismatch bugs (APR, msg insur:it) by adding aliases. One more instance of that exact pattern was missed — left inline.

david-fs-valente:
Two non-blocking notes, posting as a plain comment since they're informational rather than defects in this PR:

- `saved/route.ts` here is still the pre-transaction version (no `$transaction`, no P2002 handling) — expected, since PR #167 (transaction wrapping) hasn't merged to `dev` yet. The two PRs touch the same booth-completion block, so whichever merges second will need a straightforward rebase to reconcile.
- `src/config/index.ts`'s `constants` object now holds a single field (`actionQrCodeRefreshRateMs`) now that `actionNames` moved to `edition/actions.ts` — purely cosmetic, not worth a change on its own.

src/edition/actions.ts
dinisjcorreia: Fixed in 37667b2: added the `hitachi` alias and matching regression assertion. `pnpm test` passes 6/6 and typecheck is clean.

david-fs-valente (approved):
Re-reviewed after the Hitachi alias fix in 37667b2 — checked for other missed instances of the same pattern myself rather than trusting that one fix closed the whole class of bug.

Extracted every company display name from `DiamondCompanies.tsx`, `GoldCompanies.tsx`, `SilverCompanies.ts`, and `BronzeCompanies.ts` and cross-referenced each against `boothActions` in `src/edition/actions.ts`. Every company with a pre-existing switch case (APR, armis, hitachi, deloitte, devscope, glintt, accenture, msg insur:it) now resolves correctly via the case-insensitive lookup + aliases. Companies with no original switch case (Cloudflare, CGI, Devoteam, Cegid, Itim, Euronext, Liderteam) correctly have no map entry, matching pre-PR behavior. No further missed instances found.

Also verified:
- `pnpm test`: 6/6 pass, including the new Hitachi regression case.
- `pnpm build`: succeeds cleanly; the new `src/app/manifest.ts` correctly emits a `/manifest.webmanifest` route replacing the static `public/manifest.json` (confirmed in the build output).
- `pnpm exec tsc --noEmit`: zero new errors — remaining errors are the same pre-existing image-module-resolution issues seen elsewhere (now under `src/edition/CompaniesImages.ts` since the file was renamed, not newly broken).
- Targeted `eslint` on all touched files: clean.
- No dangling references anywhere in the tree to the old `src/utils/*` paths.
- `prisma/seed.ts`'s action list, now sourced from `edition/actions.ts`'s `actions.seed`, matches the original inline array exactly — no data drift.
- No circular dependency between `@/config` and `@/edition`.

Approving.

PR #89, opened 2025-11-25
feat: Add comment support to saved student profiles #80

samaramirandda:
Introduces an optional 'comment' field to the SavedStudent model and API routes, allowing users to add, edit, or remove comments when saving student profiles. Updates frontend components to handle comment input and display, and includes comments in the CSV export for company history.

Issue #127, opened 2026-07-12
[Refactor] Centralize request validation in src/schemas (+ length bounds)


david-fs-valente:
### Description
`src/schemas/` exists but is bypassed by inline schemas, and some inputs are unbounded.

### Why
Inline `z.object` definitions live in `students/[code]`, `user`, and `password-reset`; `name`/`bio`/`password`/`token` have no `.max()`.

### How
Move all request validation into `src/schemas/`; add `.max()` bounds on free-text fields.

### Dependencies
None.

PR #176, opened 2026-07-12
feat(saved profiles): add profile comments

### LINKED ISSUE #80: Adicionar comment quando company faz scan.
Adicionar campo no overlay com opção de comentar 
Adicionar no sumário opção de editar save para adicionar ou remover um comentário 
Adicionar comentário ao exportar 

Adicionar no savedStudent o campo comentário

dinisjcorreia:
Add optional comments to saved student profiles and expose them across save, edit, remove, history, and export flows.

Companies need context attached to saved profiles so employees can record and maintain notes while reviewing candidates. Comment updates are scoped to the logged-in company, and CSV export includes every saved student even when no CV is available. Spreadsheet-formula prefixes are escaped during CSV generation.

This also adds the Prisma migration, focused helper tests, and CI coverage for tests plus the production build. It follows up on the blockers identified in #89.

Fixes #80


src/app/api/saved/route.ts
```diff
@@ -47,6 +78,12 @@ export async function POST(req: NextRequest) {
     return NextResponse.json({ error: "Company not found" }, { status: 404 });
 
   // check if student is already saved by this company
+  if (!studentCode) {
```
david-fs-valente: [Blocking] This `if (!studentCode)` guard is placed after `prisma.student.findUnique({ where: { code: studentCode }, ... })` runs a few lines up — but `token` is optional in `saveSchema`, so a request with no `token` at all skips the decode block and reaches that `findUnique` call with `studentCode: undefined` before this guard ever executes. There's no try/catch around this section of `POST`, so that produces an unhandled error / generic 500 instead of the clean 400 this guard is meant to return.

The `PATCH` handler in this same file has the equivalent guard placed correctly — before its `isSaved()`/`findUnique()` calls — so the fix is straightforward: move this block up to right after the token-decode block (before the `student` lookup), matching `PATCH`. Not reachable via the current UI (both callers always send a `token`), but it's a real gap in a route that accepts direct API calls, and inconsistent with its own sibling handler.

david-fs-valente (changes requested):
Verified against PR head (05cbcfa) in a disposable worktree with an independently-installed node_modules (regenerated the Prisma client from this PR's schema): typecheck clean, `pnpm test` 8/8 passing. Checked open issues — this PR closes #80, and I couldn't find any issue tracking the defect below, so it's blocking here.

Good feature overall: the company-scoped comment editing is correctly IDOR-safe, the CSV formula-injection escaping is correct and tested, and the JWT-decode hardening closes a real gap. One ordering issue below, left inline.

david-fs-valente:
Non-blocking notes, posting as a plain comment since they're informational rather than defects in this PR:

- `savedStudentComments.ts`'s `csvCell` reimplements — nearly line-for-line — the CSV formula-injection escaping already in `src/app/api/export/route.ts`'s own local `csvCell`. Worth extracting one shared helper rather than maintaining two copies of injection-sensitive logic.
- Neither `saveSchema.comment` nor `savedCommentSchema.comment` has a max length, so the textarea and API accept unbounded comment text — but this matches the rest of the codebase (no schema anywhere uses `.max()`), and is exactly the gap issue #127 already tracks, so not new to this PR.
- `saved/route.ts` here is still on the pre-#167/#173 booth-switch and non-transactional version — expected, since neither has merged to `dev` yet. Same three-way overlap on this file flagged on the last two reviews; whichever of the three merges last will need a straightforward rebase.

david-fs-valente (approved):
Re-reviewed after the guard-ordering fix in 79a5f01 — verified it independently rather than trusting the commit message.

Pulled the full current `saved/route.ts`: the `if (!studentCode)` guard in `PATCH` now sits right after the token-decode block, before *both* `isSaved()` and `findUnique()` — actually a superset of what was requested (the original review only required it precede `findUnique`; the fix also guards `isSaved()`, which is safer).

Also independently checked, rather than taking the PR description's claims at face value:
- **`PUT` handler's IDOR-safety**: `savedStudentCompanyWhere(studentId, companyId)` scopes the `updateMany` by `{studentId, savedBy: {companyId}}` — correctly bounded to the caller's own company. Also confirmed `studentId` sent from the frontend (`item.studentId` in `CompanyHistorySection`) comes from `GET /api/companies/history`'s raw `SavedStudent.studentId` scalar (a UUID), matching `savedCommentSchema`'s `.uuid()` validation — no schema/payload mismatch.
- **CSV/zip data shape end-to-end**: `cv-zip/route.ts`'s query selects `student.name`/`student.code` and the default-included `comment` scalar, matching `SavedStudentCsvEntry` exactly.

Build/test verification: `pnpm test` 8/8 pass, `tsc --noEmit` zero new errors (same pre-existing baseline as other open PRs), `pnpm build` succeeds with proper env vars configured (my first attempt failed on missing Supabase env vars in my local worktree — confirmed via a side-by-side build of plain `dev` that this was a test-harness gap, not a PR defect).

Approving.

david-fs-valente:
**Non-blocking UX note:** removing `cv-zip`'s "no CVs to export" 404 and "empty zip" 500 checks is intentional and matches this PR's stated goal of always including the CSV even when no CV is available. One side effect: a company with zero saved students now silently downloads a zip containing only a CSV header row, instead of getting a clear error message. Purely a UX nicety, not a defect — not worth holding up the PR, just flagging in case it's worth a quick follow-up (e.g. a friendly empty-state message instead of a downloadable near-empty zip).

PR #183, opened 2026-07-13
refactor: Email & Tier VO #135

### LINKED ISSUE #135: [Refactor] Introduce Value Objects (Email, Tier)
### Description
Domain rules (email format, tier behavior) are duplicated as inline regex and scattered utils.

### Why
The ISEP email regex is duplicated across sign-up steps; tier -> color/href logic is spread across several utils.

### How
Add a `domain/` layer with Email and Tier first (branded types for invariant-only cases, classes where there is behavior). Repositories convert entity <-> VO; DTOs convert VO -> primitive.

### Dependencies
None (pairs with the data-layer cleanup and DTO issues).

Pedro-Sousa7:
- Added Email VO (branded type) 
- Used Email VO across the codebase to ensure type-safety
- Added Tier VO, and implemented it across the codebase
- Unit tests


dinisjcorreia (changes requested):
Blocking regression in `CompaniesContainer`: it receives `tier`, but renders `<Company {...company} />` without passing it. None of the current company data entries define their own tier, so every card falls back to `BRONZE`. For Diamond and Gold entries, `hrefByCompanyTier` then returns `/` because those entries have no `websiteUrl`, breaking all internal company-page links.

Pass the container tier to each card (for example, `<Company {...company} tier={tier} />`) and add a check covering the container-to-card wiring.

Also resolve the current conflicts with `dev` before merge.

PR #166, opened 2026-07-12
docs: add AGENTS.md and CLAUDE.md

david-fs-valente:
## Summary
- Adds AGENTS.md covering contribution workflow (branch/commit conventions), stack, architecture layout, data model, auth model, coding conventions, a verification/definition-of-done checklist, and known gotchas.
- Adds a thin CLAUDE.md that imports AGENTS.md, so it stays the single source of truth across tools.

closes #125

PR #185, opened 2026-07-13
test: establish Vitest testing foundation

### LINKED ISSUE #148: [Test] Establish a testing foundation (Vitest)
### Description
There are zero tests, so every refactor ships blind.

### Why
No test tooling or test files exist.

### How
Add Vitest + Testing Library; write tests against services as they are extracted. Do not gate on coverage initially.

### Dependencies
Depends on #126 — meaningful unit tests target the extracted services (test setup alone is independent).

dinisjcorreia:
## Summary

- replace node:test runner with Vitest and migrate all discovered tests
- configure jsdom, Testing Library cleanup, and jest-dom matchers
- add service orchestration and component smoke tests
- run tests in pull-request CI without coverage gating
- document Vitest auto-discovery and CI behavior

## Verification

- pnpm install --frozen-lockfile
- pnpm test (31 passed)
- pnpm lint
- pnpm typecheck

Closes #148
Depends on completed #126


david-fs-valente:
**Heads-up (non-blocking): AGENTS.md is now stale**

This PR changes `pnpm test` from the explicit `tsx --test <8 files>` list to `vitest run`, and wires it into CI for the first time — but `AGENTS.md` still documents the old command, the old "eight files" count, and states CI doesn't run `pnpm test`. Not blocking here since AGENTS.md is already being reworked in #166.

Suggested sequencing: land #166 first, then merge `dev` into this branch so the AGENTS.md doc updates land together with (or ahead of) this test-runner swap, instead of drifting further out of sync.

david-fs-valente (changes requested):
Requesting changes on two points found while reviewing the Vitest migration:

1. **`src/application/boundaries.test.ts` — asymmetric fix.** The new `if (/\.test\.tsx?$/.test(file)) continue;` guard is added only to the server-folder loop (`application/repositories`, `application/services`, `lib/http`), not to the `client/` loop right below it. There's no `client/*.test.ts` file yet so it doesn't fail today, but the same problem will hit as soon as a client-side test is co-located under `src/client/` (no `import "client-only";` in a test file). Please apply the same skip to the client loop for consistency.

2. **`vitest.config.ts` only aliases `@/*`, not `~/*`.** `tsconfig.json` defines both (`@/*` → `./src/*`, `~/*` → `./*`), but only `@/*` is aliased in the Vitest config's `resolve.alias`. Four files currently use `~/` (`components/Map/index.tsx`, `app/page.tsx`, `app/(profiles)/student/[...data]/page.tsx`, `edition/CompaniesImages.ts`) — none are under test today, so this doesn't fail now, but the first test written against one of them will hit a confusing module-resolution error instead of an obvious alias mismatch. Please add the `~/*` alias alongside `@/*`.

david-fs-valente:
**Minor / non-blocking suggestions**

- `test.environment` is set globally to `"jsdom"` for all test files, including pure-logic/filesystem ones (`boundaries.test.ts`, `authFlow.test.ts`, `serviceLogic.test.ts`, etc.) that don't touch the DOM. Fine at the current 10-file suite size; if it grows, consider `environmentMatchGlobs` or per-file `// @vitest-environment node` pragmas to keep non-UI tests fast.
- Coverage gating is explicitly out of scope per the PR description — reasonable to land incrementally, just flagging it's not forgotten.

david-fs-valente (changes requested):
## Requesting changes

**Merging this as-is will break `pnpm test` (and the new CI "Test" step).**

`dev` has moved since this branch was cut: PR #183 merged as `1bad6d5` and added three test files that still use `node:test`:
- `src/types/Email.test.ts`
- `src/domain/Company/company-tier.test.ts`
- `src/components/Companies/CompaniesContainer/CompaniesContainer.test.tsx`

This PR switches `package.json`'s `test` script from an explicit file list to `vitest run`, with no `test.include`/`exclude` override in `vitest.config.ts` — so Vitest auto-discovers every `*.test.ts(x)` file, including those three.

I reproduced this in isolation: a file that does `import test from "node:test"` and calls `test(...)` runs via Node's own test runner (prints its own pass/fail independently) but registers **zero** tests with Vitest's collector, and Vitest hard-fails the suite:

```
✔ dummy legacy test (1.878644ms)
 ❯ pkg-test.test.ts (0 test)
FAIL  pkg-test.test.ts
Error: No test suite found in file ...
Test Files  1 failed (1)
```

So after rebasing onto current `dev`, `pnpm test` (and CI) goes red on files this PR never touched. The PR description's "21 passed" verification note was almost certainly run against the older base commit, before #183 landed.

**Suggested fix:** rebase onto current `dev` and migrate those three files to Vitest (`test`/`expect` from `"vitest"`), since that's consistent with the direction of this PR anyway. An explicit `test.include`/`exclude` allowlist would sidestep it but undercuts the point of switching to auto-discovery.

david-fs-valente:
A couple of minor, non-blocking items from review:

- **Import alias nit** — `src/components/HeadingText/index.test.tsx` imports via `~/src/components/HeadingText`, but the rest of the codebase consistently uses `@/components/...` (the `@` alias already maps to `./src`). The `~` alias only got added to `vitest.config.ts` to support this one import; `@/components/HeadingText` would work identically and match existing convention.

- **Docs drift** — `AGENTS.md`'s description of `pnpm test` (the "runs eight files" gotcha, and the literal command under "Common commands") is now stale: it was already out of date after #183 added 3 more test files, and this PR changes the approach entirely (explicit list → auto-discovery). Not blocking for this PR, but worth a follow-up doc pass once the `node:test` migration above is sorted.

Everything else looks good — the new `userService.test.ts` and `HeadingText` smoke test are solid, the `boundaries.test.ts` exclusion for `.test.ts(x)` files is correctly scoped, and the CI/config wiring is clean.

david-fs-valente (approved):
Re-reviewed after the fixes (commits `test: address Vitest review feedback` and `test: migrate newly added tests to Vitest`).

✅ **Previous blocking issue confirmed fixed.** I checked out the PR branch in an isolated worktree, ran `pnpm install --frozen-lockfile && pnpm test`: **13 test files / 31 tests pass**, including the three files (`src/types/Email.test.ts`, `src/domain/Company/company-tier.test.ts`, `src/components/Companies/CompaniesContainer/CompaniesContainer.test.tsx`) that were previously stuck on `node:test` and would've broken Vitest's auto-discovery. `pnpm typecheck` is also clean. `AGENTS.md` was correctly updated to describe the new auto-discovery behavior instead of the stale "eight files" list.

✅ The `~/src/components/...` import nit from the earlier pass is also fixed — `HeadingText/index.test.tsx` now uses `@/components/HeadingText`, matching repo convention.

⚠️ **One thing to sort out before merging (not a code issue):** GitHub currently reports this PR as **CONFLICTING** against `dev`. A same-day commit (`d7964b8`, "chore: remove build-time error and lint suppression") edited the same `AGENTS.md` paragraphs this PR touches (the `strict: true` gotcha, the Verification/CI paragraph, and the "pnpm test runs eight files" bullet). It's a textual overlap, not a semantic conflict — both sets of edits are compatible — so a rebase should resolve it cleanly.

Approving on the code itself.

david-fs-valente:
One more tiny nit (non-blocking): \`vitest.config.ts\`'s \`resolve.alias\` still defines a \`~\` alias pointing at the project root, but nothing imports through it anymore now that \`HeadingText/index.test.tsx\` uses \`@/components/HeadingText\`. Could drop it — purely cosmetic, not worth blocking on.

david-fs-valente:
Rebased onto current \`dev\` to resolve the \`AGENTS.md\` conflict with \`d7964b8\` — kept dev's now-accurate statement that \`next build\` fails on type/lint errors (the \`ignoreBuildErrors\`/\`ignoreDuringBuilds\` flags were removed there) while keeping this PR's addition that CI now also runs \`pnpm test\`; dropped the stale "eight files" Gotchas bullet in favor of the Vitest auto-discovery one.

Also addressed the non-blocking nit: removed the unused \`~\` alias from \`vitest.config.ts\`.

Verified post-rebase: \`pnpm install --frozen-lockfile\`, \`pnpm test\` (13 files / 31 tests passing), \`pnpm typecheck\`, \`pnpm lint\` (0 errors) all clean.

PR #187, opened 2026-07-13
chore: remove build-time error and lint suppression

### LINKED ISSUE #150: [Chore] Remove build-time error/lint suppression
### Description
Type and lint errors ship to production regardless of any gate.

### Why
`next.config.js` sets `typescript.ignoreBuildErrors: true` and `eslint.ignoreDuringBuilds: true`.

### How
Remove both once strict mode and the CI gate are in place.

### Dependencies
Depends on #124 (CI gate) and #128 (strict mode) — removing the escape hatch is only safe once they catch what the build would.

dinisjcorreia:
Closes #150

## Summary
- remove TypeScript build-error suppression
- remove ESLint build suppression
- let Next.js production builds fail on type or lint errors

## Verification
- `pnpm typecheck`
- `pnpm lint`
- `NEXT_PUBLIC_SUPABASE_URL=https://example.supabase.co NEXT_PUBLIC_SUPABASE_ANON_KEY=test-anon-key pnpm build`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,7 +28,7 @@ For every requested task:
 | Layer | Tech |
 |---|---|
 | Framework | Next.js 15 (App Router), React 18 |
-| Language | TypeScript (`strict: true` — see [Gotchas](#gotchas) for what still slips through anyway) |
+| Language | TypeScript (`strict: true`) |
 | Styling | Tailwind CSS 4, HeroUI 2.8 |
 | Database | PostgreSQL via Supabase, Prisma 6 (`prisma/schema.prisma`) |
 | Auth | Supabase Auth (session) — see [Auth model](#auth-model) |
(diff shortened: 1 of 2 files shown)


david-fs-valente (approved):
Verified locally on the branch: \`pnpm typecheck\` and \`pnpm lint\` both clean, and reproduced the PR's own verification step — \`NEXT_PUBLIC_SUPABASE_URL=... NEXT_PUBLIC_SUPABASE_ANON_KEY=... pnpm build\` succeeds with exit code 0 (37/37 static pages) now that the suppression is removed. Also grepped for any other reference to \`ignoreBuildErrors\`/\`ignoreDuringBuilds\` (Dockerfile, docs, configs) — none found, so this is a complete, self-contained removal.

Small, correct, verified-working fix for #150. One non-blocking sequencing note left as a separate comment.

david-fs-valente:
**Minor (non-blocking): future collision with #166's AGENTS.md**

Not a problem with this diff — but worth flagging for sequencing. \`AGENTS.md\` (currently only on the still-open \`docs/agent-docs\` PR #166, not yet on \`dev\`) has a Gotchas bullet stating \`next build\` "will not fail on a type or lint error." Once both #166 and this PR land, whichever merges second will leave that bullet inaccurate and should drop it. Nothing to do here since \`AGENTS.md\` doesn't exist in its full form on \`dev\` yet — just flagging the future collision so it doesn't get missed.

dinisjcorreia:
Addressed now that #166 has merged. Rebased this branch onto current `dev`, removed the stale Gotchas bullet, and corrected the related Stack and Verification references so `AGENTS.md` now reflects that `next build` fails on type/lint errors.

Verification on the rebased head: `pnpm typecheck`, `pnpm test` (28/28), `pnpm lint` (0 errors; 13 pre-existing warnings from `dev`), and the production build (37/37 static pages) all pass.

PR #188, opened 2026-07-13
fix: enforce company-scoped uniqueness on SavedStudent

### LINKED ISSUE #151: [Fix] Fix SavedStudent uniqueness grain (prevent duplicate company saves)

david-fs-valente:
## Summary
- `SavedStudent`'s primary key is `(studentId, employeeId)` (employee-scoped), but the dedup rule the app enforces (`isStudentSaved`) is company-scoped, so two employees at the same company could race and both save the same student — no DB backstop.
- Adds a `companyId` column to `SavedStudent` with a `@@unique([studentId, companyId])` constraint, so the database now enforces the same grain the app assumes.
- Migration backfills `companyId` from the saving employee, and collapses any pre-existing duplicate company saves (keeping the earliest one) before adding the constraint.
- Repository queries (`countCompanySaves`, `findCompanyHistory`, etc.) now filter on the direct `companyId` column instead of joining through `savedBy.companyId`.

Closes #151

### DIFF
--- a/prisma/migrations/20260713000000_saved_student_company_scope/migration.sql
+++ b/prisma/migrations/20260713000000_saved_student_company_scope/migration.sql
@@ -1,0 +1,26 @@
+-- AlterTable: add companyId as nullable so it can be backfilled first
+ALTER TABLE "SavedStudent" ADD COLUMN "companyId" UUID;
+
+-- Backfill companyId from the saving employee's company
+UPDATE "SavedStudent" ss
+SET "companyId" = e."companyId"
+FROM "Employee" e
+WHERE e."id" = ss."employeeId";
+
+-- Collapse pre-existing duplicate company saves (same student saved by more
+-- than one employee of the same company), keeping the earliest save per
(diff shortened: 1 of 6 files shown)


dinisjcorreia (approved):
Reviewed migration and all SavedStudent write/query paths. Company-scoped uniqueness, backfill, duplicate cleanup, and repository updates align with #151.\n\nValidated: pnpm test (28 passing), pnpm typecheck. pnpm lint has no errors; existing warnings are outside this PR.

PR #189, opened 2026-07-13
refactor: add error/loading handling to client mutations

david-fs-valente:
## Summary
Closes #142.

Client `fetch` mutations across ~10 components had no error handling
and/or no pending state, so a failed request either left the UI stuck
(a permanently-loading button) or failed silently with no user feedback.

- Add `useMutation` — a small shared hook (pending state + `toast.error` on thrown errors) that mutation handlers wrap their body in.
- Add `useLogout`, consolidating three copies of the same confirm+fetch+clear-session logout logic (`LogoutButton`,
  `CompanyProfileSectionContainer`, `ProfileSectionContainer`) that had no error handling if the logout request failed.
- Wire the hook into the components with real gaps: `AdminSavedSection` (the bug named in the issue), `CloseActionButton`, `CompanyStatsSection` (had a dead `isLoading` state that was never turned on), `CompanyViewProfileSectionContainer`, `PreviewProfileSectionContainer`, `ProfileSection`/`SettingsSection` (could get stuck in a loading state forever on a network error — `SettingsSection`'s error branch also showed a false success message), `QRCodeTab/CompanyTab`, `QRCodeTab/ScanTab`, and `CompanySavedProfilesSection.handleManualEntry`.

Components that already had solid try/catch + pending-state handling (e.g. `CompanyHistorySection`, most of `CompanySavedProfilesSection`, `QRCodeTab/ScanTab`'s scan handler) were left alone.

PR #190, opened 2026-07-13
refactor: extract an HTTP layer

### LINKED ISSUE #149: [Refactor] Extract an HTTP layer (client/server split, HTTP-agnostic services)
### Description
Client HTTP calls are scattered raw `fetch`, both axios and fetch are used, and handlers repeat auth -> parse -> respond inline.

### Why
This is the top candidate for a reusable framework package and the structural fix for missing per-route authorization.

### How
Add `lib/http/client.ts` (HttpClient + request builder) and `lib/http/server.ts` (`defineHandler` with a pluggable auth strategy). Services stay HTTP-agnostic. Unify on one transport first.

### Dependencies
Depends on #126 — services must exist to be wrapped and kept HTTP-agnostic.

david-fs-valente:
## Summary

Closes #149. Client calls were raw `fetch(BASE_URL + ...)` scattered everywhere (plus a dead `axios` dep), and routes repeated the same auth → parse → respond → error-map by hand.

- **`lib/http/server.ts`** — `defineHandler({ auth, schema?, authorize?, handler })` with a pluggable auth Strategy (`public`/`session`/`student`/`employee`/`admin`) and an `authorize()` hook for ownership checks.
- **`lib/http/client.ts`** — `httpClient`, a typed fetch wrapper that throws `HttpClientError` on non-2xx, plus a `raw()` escape hatch for blob/redirect responses.
- **22/27 routes** migrated to `defineHandler`. The other 5 stay raw with an inline comment explaining why (health check, JWT-in-query export route, two form-data upload routes).
- **All client fetch call sites** migrated to `httpClient`.
- Removed unused `axios`.
- `AGENTS.md` updated to document the new pattern.

A few already-inconsistent routes had their auth-failure status/shape harmonized to match the rest — none of the current clients branch on the old specific codes, so these are cleanups, not breaking changes.

**Note:** cut from `dev` before #189 merged, so a few shared files will conflict with that PR — expected, not reconciled here.


david-fs-valente:
Closing in favor of a version stacked on #189 (refactor/client-mutation-error-handling) — this branch was cut from dev before #189 and conflicts with it on ~12 files that both PRs touch. Replacement PR incoming targeting that branch instead.

PR #191, opened 2026-07-14
refactor: prisma schema hygiene (UUID typing, indexes, timestamps, enums)

### LINKED ISSUE #155: [Refactor] Prisma schema hygiene (UUID typing, indexes, timestamps, enums)
### Description
Several small correctness/consistency issues in the Prisma schema.

### Why
Inconsistent UUID typing (Action/ActionCompletion are text vs `@db.Uuid`); redundant indexes (Student.code, Company.name already unique); missing `createdAt` on core entities; the Role enum vs a written `"COMPANY"`; `Student.year` is a free string; avatar/cv have no integrity.

### How
Standardize `@db.Uuid`; drop redundant indexes; add `createdAt`; clarify the role model; consider a year enum and an Upload table. Ship as migrations.

### Dependencies
Depends on #138 — all schema changes ship as migrations.

david-fs-valente:
## Summary

Closes #155.

- **UUID typing** - `Action.id`, `ActionCompletion.id`, `ActionCompletion.actionId` were stored as `text`; standardized to `@db.Uuid` like every other model.
- **Redundant indexes** - dropped `Student.@@index([code])` and `Company.@@index([name])`; both already have `@unique`, which creates its own index.
- **Timestamps** - added `createdAt` to `User`/`Student`/`Company`/`Employee` (they only had `updatedAt`).
- **`Student.year` enum** - was a free `String` validated only in zod. Added a `Year` enum via `src/domain/Student/year.ts` (mirrors the existing `domain/Company/company-tier.ts` value-object pattern). Prisma-side identifiers are decoupled from the DB-stored label via `@map`, so no data backfill is needed. The conversion lives at the two seams that touch Prisma (`studentRepository`/`studentService` on write, `dto/studentDto.ts` on read) - nothing else changes, so zod validation, the sign-up dropdown, and every display component keep working with the same label strings as before.
- Documented (schema comment, not fixed) that `avatar`/`cv` are bare file-IDs with no DB integrity - that's #37's job, not this hygiene pass's.
- The "Role enum vs `COMPANY`" item from the plan was already resolved separately.

### Migration safety
Prisma's auto-generated plan for the UUID/enum type changes would have **dropped and recreated** those columns - fine on an empty table, but data loss on any table with real rows. I hand-edited the migration to use `ALTER COLUMN ... TYPE ... USING (...)` casts instead, and verified it against a throwaway Postgres container seeded with representative rows (an `Action`, an `ActionCompletion`, a `Student` with a year value): all IDs and the year value survived the migration intact.


dinisjcorreia (approved):
Reviewed migration and Student year mapping. No blocking issues found. Verified: pnpm test (31 passing), pnpm lint (existing warnings only), pnpm typecheck, and Prisma schema validation.

Issue #26, opened 2025-10-05
[Task] Update Prisma Seed


david-fs-valente:
Update the seed of Prisma to the new database.

PR #192, opened 2026-07-14
refactor: extract an HTTP layer - client/server split, HTTP-agnostic services

### LINKED ISSUE #149: [Refactor] Extract an HTTP layer (client/server split, HTTP-agnostic services)

david-fs-valente:
## Summary

Closes #149. Stacked on #189 (refactor/client-mutation-error-handling) — the original attempt at this (#190) was cut from `dev` before #189 and conflicted with it on ~12 files both PRs touch, so this one is rebuilt on top of #189 instead. **Base should be retargeted to `dev` once #189 merges.**

Client HTTP calls were scattered raw `fetch(BASE_URL + ...)`, both `axios` and `fetch` were pulled in (axios turned out to be fully dead — zero imports), and route handlers repeated the same auth → parse → work → respond → error-map inline.

- **`lib/http/server.ts`** (`"server-only"`): `defineHandler({ auth, schema?, authorize?, handler })`. `auth` is a Strategy (`"public" | "session" | "student" | "employee" | "admin"`, default `"session"`) checked by a pure predicate, `application/domain/authPolicy.ts`'s `passesAuthPolicy` (unit tested directly). `authorize` is an optional `(session, params) => boolean` for ownership checks that runs after the auth Strategy passes.
- **`lib/http/client.ts`** (`"client-only"`): `httpClient`, a `FetchHttpClient` behind an `HttpClient` interface (`get`/`post`/`patch`/`put`/`delete`, JSON in/out, throws a typed `HttpClientError` with `.status` on non-2xx) plus a `raw()` escape hatch for blob/redirect responses.
- **22 of 27 API routes** migrated to `defineHandler`. The other 5 stay raw with an inline comment explaining why (health check, query-JWT export route, two form-data upload routes with a pre-existing unrelated auth gap tracked as #26).
- **Every client fetch call site** migrated to `httpClient` — both the ~10 components #189 doesn't touch, and the ~12 it does (where `useMutation`/`useLogout` now call `httpClient` internally instead of raw `fetch`, catching `HttpClientError` wherever the code branches on status like 409/400).
- Removed the now-unused `axios` dependency.
- `AGENTS.md` updated to document the new pattern.


dinisjcorreia (commented):
One non-blocking scope gap:

`src/app/(auth)/password-reset/page.tsx` is still a Client Component using raw `fetch` for `/api/auth/password-reset`. This bypasses the new `httpClient` error/status handling, despite the PR description and updated guidance saying client fetch call sites were migrated. Please move this call to `httpClient.post` (or narrow the claim/guidance) so the HTTP layer is consistently adopted.

Verification: `pnpm test` (40 passed) and `pnpm typecheck` passed. `pnpm lint` has four existing warnings in `src/components/Companies/CompaniesSection/index.tsx`, outside this PR.

dinisjcorreia (approved):

david-fs-valente:
@dinisjcorreia Good catch — fixed in 24a0bc0: `password-reset/page.tsx` now goes through `httpClient.post` instead of raw `fetch`, so it picks up the same `HttpClientError`/status handling as the rest of the client call sites.

PR #193, opened 2026-07-14
refactor: move company display data into the Company table

### LINKED ISSUE #152: [Refactor] Move company display data into the Company table
### Description
There are two sources of truth for "company": the DB table and hardcoded utils.

### Why
Public pages read `utils/*Companies`, while the DB table feeds recruitment; admins cannot toggle active/rank without a deploy.

### How
Extend `Company` with `active` and (as needed) website/order; public pages read from the DB. Store logos in Storage and the URL in the column.

### Dependencies
Depends on #138 — schema change, ships as a migration.

david-fs-valente:
## Summary

Closes #152. Stacked on #192 (refactor/http-client-server-layer-v2) so the new admin CRUD routes use `defineHandler`/`httpClient` from day one instead of the old inline-auth pattern. **Base should be retargeted to `dev` once the #189 → #192 stack merges.**

Two sources of truth for "company" existed: the DB table (recruitment/auth side) and hardcoded `edition/*Companies.tsx` (public pages). Admins couldn't toggle a company active or fix the roster without a deploy.

- **Schema**: `Company` gains `website`, `active` (defaults `false` - self-registered companies stay hidden from the public site until an admin vets them), `order`. Migration backfills the 15 existing companies from `edition/*Companies.tsx`, upserted by name so an already-self-registered company (with a real recruiting account) gets its display fields filled in rather than duplicated - verified against a throwaway Postgres container seeded with a pre-existing company+employee to confirm the upsert preserves IDs/FKs instead of Prisma's default (unsafe on non-empty tables) drop-and-recreate plan.
- **Scope call**: per discussion, only roster fields (name/logo/website/active/order) move to the DB. The rich per-company modal content (JSX body text, social links, video, facts with icon components) isn't safely DB-editable as plain content, so it stays in `edition/*Companies.tsx`, now looked up by name (`findEditionContentByName`) instead of being the roster source of truth.
- **New routes**: `GET /api/companies/roster` (public, active companies ordered) and `GET`/`POST /api/admin/companies` + `PATCH /api/admin/companies/[id]` (admin CRUD).
- **New admin page** (`/companies`) to create/edit companies and toggle `active` - the actual capability the issue asked for.
- **Public pages rewired**: `CompaniesSection` fetches the roster and merges in edition content client-side; `company/[name]/page.tsx` (already a Server Component) fetches the DB row directly and passes it down.
- `Company`'s `logoHref` changes from a Next.js static image import to a plain URL string; logo dimensions (needed now that there's no static import to infer them from) were measured from the actual asset files via `sharp` rather than guessed.


dinisjcorreia (changes requested):
Blocking: PR is currently `CONFLICTING` with its target branch. Rebase onto `refactor/http-client-server-layer-v2` and resolve the conflicts before merge. In particular, retain base branch's `Company.createdAt` field alongside these new display fields; also resolve the `CompaniesContainer` test import conflict.

src/components/Companies/CompaniesSection/index.tsx
```diff
@@ -1,18 +1,68 @@
...
+import type { CompanyRosterDto } from "@/application/dto/companyDto";
+import { COMPANY_TIER, CompanyTier } from "@/domain/Company/company-tier";
+import { findEditionContentByName } from "@/edition";
+
+function toCompanyProps(company: CompanyRosterDto): CompanyProps | null {
+  if (!company.avatar) return null;
+
+  const editionContent = findEditionContentByName(company.name);
+
+  return {
```
dinisjcorreia, line 17: New active DIAMOND/GOLD companies are rendered here even when they have no edition content. Their card is routed to `/company/<name>` by `hrefByCompanyTier`, but that page returns `Custom404()` when `findEditionContentByName()` has no modal data. The new admin UI permits this state (create company, add logo/website, set active), so admins can publish a card that always 404s. Fall back to `website` when no edition content exists, or prevent activation for these tiers until content exists.

david-fs-valente:
@dinisjcorreia Addressed both:

- **Merge conflict** (021eeb6): rebuilt on top of the latest `refactor/http-client-server-layer-v2`. In `prisma/schema.prisma`, kept both sides — base's `Company.createdAt` alongside this PR's `website`/`active`/`order` (the two additive migrations already apply in the right order, no migration changes needed). Also resolved a `companyDto.ts` conflict (dropped the now-orphaned `InterestMatchDto` the base branch had already deleted) and the `CompaniesContainer` test conflict (moved to Vitest per base, kept this branch's plain-string `logoHref`). PR is now `MERGEABLE`.
- **404 on content-less DIAMOND/GOLD cards** (5743178): went with the "fall back to website" option. `hrefByCompanyTier` now takes a `hasContent` flag and only routes to the internal `/company/[name]` page when the company has `modalInformation` to show there — otherwise it falls back to `websiteUrl || "/"`, same as SILVER/BRONZE. `Company` passes `Boolean(modalInformation)` through. Updated both `company-tier.test.ts` and `CompaniesContainer.test.tsx` for the new signature.

`pnpm typecheck`, `pnpm lint`, and `pnpm test` (41/41) all green after the rebase and fix.

dinisjcorreia (approved):
Re-reviewed latest head. Merge conflicts resolved, `createdAt` retained, and content-less DIAMOND/GOLD cards now fall back to their website instead of 404ing. Approved.

PR #194, opened 2026-07-14
refactor: move sponsors into the DB

### LINKED ISSUE #153: [Refactor] Move sponsors into the DB or edition config
### Description
Sponsors are hardcoded, with the same shape as companies.

### Why
`utils/Sponsors.ts` holds them in code, so changes need a deploy.

### How
If sponsors are edited live by non-devs, use a `Sponsor` table with admin CRUD; if set once per edition, put them in the edition config. Do not build an admin surface just for sponsors.

### Dependencies
Depends on #129 (edition-config path) — or on the company-in-DB work if going the table route.

david-fs-valente:
## Summary

Closes #153. Stacked on #193 (refactor/company-table-display), reusing its admin-CRUD pattern directly. **Base should be retargeted to `dev` once the #189 → #192 → #193 stack merges.**

Per our discussion: sponsors change edition-to-edition (this year's RedBull may not be next year's) and admins need a way to fix a forgotten one without a deploy. Since #193 was already building the Company admin-CRUD mechanism, the marginal cost of a parallel `Sponsor` table is small - the plan's own carve-out for this ("fold sponsors into the DB+admin mechanism only if you're already building company CRUD").

- **Schema**: new `Sponsor` table (`name`, `logo`, `website`, `active`, `order`). Unlike `Company`, sponsors have no self-registration path, so `active` defaults to `true` - there's no untrusted signup to vet.
- Migration backfills the 7 existing sponsors from `edition/Sponsors.ts`, referencing their existing static asset paths (no re-upload).
- **New routes**: `GET /api/sponsors` (public) and `GET`/`POST /api/admin/sponsors` + `PATCH /api/admin/sponsors/[id]` (admin CRUD), mirroring the company routes.
- **New admin page** (`/sponsors`) to create/edit sponsors.
- `SponsorsSection` now fetches from the DB instead of the static `edition.sponsors` array; `Sponsor`'s `logoHref` changes from a static image import to a plain URL string (switched to `fill`+`object-contain` inside a sized wrapper, matching the old 85%/90% visual cushion).
- Removed `edition/Sponsors.ts` and `edition/CompaniesImages.ts` (fully unused once both #193 and this land) and the now-dead `tiers`/`sponsors` keys on the `edition` object.
- **Bug fix carried into #193 too**: found that `adminCompanySchema.ts`'s `avatar` field used `z.url()`, which rejects the relative `/assets/images/companies/...` paths the migration backfill stores - saving a pre-existing company row without touching its logo would fail validation. Fixed there and applied the same fix here for `adminSponsorSchema.ts`'s `logo` field up front.


src/schemas/adminSponsorSchema.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// Logos may be an absolute URL (a freshly-uploaded Storage object) or an
+// absolute path into public/ (the pre-existing static sponsor assets - see
+// the add_sponsor_table migration's backfill, which references them by
+// path rather than re-uploading them).
+const logoSchema = z
+  .string()
+  .max(2048)
+  .refine(
+    (value) => value.startsWith("/") || z.url().safeParse(value).success,
```
dinisjcorreia: This accepts every absolute URL, but `Sponsor` passes the value to `next/image` and `next.config.js` only allows Supabase, GCS, and Firebase hosts. An admin can create a sponsor with a normal external URL (for example, `https://example.com/logo.png`); the POST succeeds, then the public sponsor section throws Next’s “hostname is not configured” error when rendering it. Restrict logo inputs to configured storage origins (or serve remote logos without `next/image`) before exposing this admin flow.

dinisjcorreia (changes requested):
One blocking issue found.

src/schemas/adminSponsorSchema.ts
david-fs-valente: Fixed — extracted a shared `logoSchema` (`src/schemas/logoSchema.ts`) that restricts absolute logo URLs to the hosts `next.config.js`'s `images.remotePatterns` actually allows (Supabase, GCS, Firebase, plus the local Supabase gateway in dev), instead of accepting any absolute URL. Applied the same fix to `adminCompanySchema.ts`, which had the identical pattern. Also retargeted this PR to `dev` now that #189→#192→#193 have merged.

src/schemas/logoSchema.ts
```diff
@@ -0,0 +1,44 @@
...
+    ...(serverEnv.NODE_ENV !== "production" ? [/^127\.0\.0\.1$/] : []),
+  ];
+}
+
+function isAllowedLogoUrl(value: string): boolean {
+  const parsed = z.url().safeParse(value);
+  if (!parsed.success) return false;
+
+  const { hostname } = new URL(value);
+  return allowedLogoHostnames().some((pattern) => pattern.test(hostname));
```
dinisjcorreia: [P1] Match the complete `remotePatterns` rule

This still accepts URLs that `next/image` rejects because it checks only the hostname. For example, `https://demo.supabase.co/not-storage/logo.png` passes this schema, but `next.config.js` requires `/storage/v1/object/public/**`; HTTP Supabase URLs and wrong local ports also pass here and then fail during rendering. The `value.startsWith("/")` branch additionally accepts protocol-relative values such as `//example.com/logo.png`, which Next rejects. Validate protocol, port, and pathname against the same patterns (ideally through one shared matcher) so the API cannot persist a logo that breaks the public section.

src/schemas/adminSponsorSchema.ts
```diff
@@ -0,0 +1,24 @@
+import { z } from "zod";
+
+import { logoSchema } from "@/schemas/logoSchema";
+
+export const createAdminSponsorSchema = z.object({
+  name: z.string().min(1).max(100),
+  logo: logoSchema,
+  website: z.url().max(2048).nullable().optional(),
```
dinisjcorreia: [P1] Restrict public website links to HTTP(S)

`z.url()` accepts `javascript:`, `data:`, and other non-web schemes. This value is persisted and rendered directly as the public Sponsor `<a href>`, and React 18 leaves a `javascript:` URL in the generated markup. Accept only `http:`/`https:` URLs here; the identical company website fields should use the same shared schema.

src/components/SponsorsSection/index.tsx
```diff
@@ -1,14 +1,32 @@
...
+  if (!sponsors) return null;
+
   return (
     <section className="flex flex-col items-center gap-y-10 border-b border-b-secondary bg-background text-center">
-      <SponsorsContainer sponsors={edition.sponsors} />
+      <SponsorsContainer
+        sponsors={sponsors.map(({ name, logo, website }) => ({
+          name,
+          logoHref: logo,
+          website: website ?? "",
```
dinisjcorreia: [P2] Do not turn a missing website into a current-page link

`website` is nullable and the create form does not ask for it, so each newly created sponsor initially reaches this branch. Mapping `null` to `""` still makes Sponsor render an anchor; clicking its logo opens the current page in a new tab. Preserve the optional value and render a non-link logo when absent, or require a website during creation.

dinisjcorreia (changes requested):
Three blocking issues remain. The prior logo-rendering blocker is only partially fixed; website handling also needs tightening before this admin-managed data is published.

src/schemas/logoSchema.ts
david-fs-valente: Fixed — extracted the pattern list into `src/config/imageRemotePatterns.js` as a single source of truth for both `next.config.js` and this schema, and rewrote the matcher to check protocol, port, hostname, *and* pathname (matching `https://demo.supabase.co/not-storage/...` no longer passes). Also closed the protocol-relative (`//host/...`) bypass of the leading-"/" path check. Added `src/schemas/logoSchema.test.ts` covering these cases.

src/schemas/adminSponsorSchema.ts
david-fs-valente: Fixed — extracted a shared `websiteUrlSchema` (`src/schemas/websiteUrlSchema.ts`) restricted to `http`/`https` via `z.url({ protocol: /^https?$/ })`, and applied it to both `adminSponsorSchema.ts` and `adminCompanySchema.ts`'s website fields. Added a test covering `javascript:`/`data:` rejection.

src/components/SponsorsSection/index.tsx
david-fs-valente: Fixed — `Sponsor`'s `website` prop is now optional, and the component renders a plain, non-clickable logo (no `<a>` wrapper) when it's absent instead of falling back to an empty-string href. `SponsorsSection` now maps a null website to `undefined` instead of `""`.

dinisjcorreia (approved):
Reviewed current head. Previous blocking findings are addressed; no blocking issues found.

PR #195, opened 2026-07-14
chore(deps): remove unused swal package

### LINKED ISSUE #131: [Chore] Dependency cleanup (redundant alert libs, duplicate framer-motion)
### Description
Redundant and duplicated dependencies remain in `package.json`.

### Why
`swal` and `sweetalert` are both present (two alert libs); `framer-motion` is declared twice (`^12` and `^10`). (firebase-admin was already removed.)

### How
Drop one alert library; pin `framer-motion` to a single version. Document the `jsonwebtoken` + Supabase boundary in the baseline.

### Dependencies
None (overlaps the notification-library consolidation issue).

david-fs-valente:
## Summary

Closes #131.

- `swal` and `sweetalert` were both present as dependencies, but `swal` has zero imports anywhere in `src` — every `import swal from "..."` actually resolves to the `sweetalert` package (same variable name, different package). Removed the dead `swal` dependency.
- The other two items in the issue were already resolved on `dev` before this PR: the `framer-motion` duplicate declaration was fixed in an earlier commit, and the `jsonwebtoken` + Supabase auth boundary is already documented in `AGENTS.md`'s Auth model section.


dinisjcorreia (approved):
Reviewed dependency removal. No blocking findings: no source references to `swal`; frozen install, tests, and typecheck pass.

PR #197, opened 2026-07-14
refactor: replace array-index keys with stable identity

### LINKED ISSUE #143: [Chore] React hygiene: stable keys and shared-type placement
### Description
Minor correctness and robustness smells in the components.

### Why
Six uses of array-index-as-`key`; a client component imports a type from a Prisma-backed module (fragile boundary).

### How
Use stable keys; move shared types out of server modules into `src/types/` (or the DTO layer).

### Dependencies
None.

david-fs-valente:
## Summary

Closes #143.

- `FactSection`: `key={i}` → `key={fact.description}`
- `InputSelect`: `key={index}` → `key={option}` (options are unique strings)
- `QRCodeTab/index.tsx`: `key={index}` → `key={title}` (tab titles are unique; `index` is still used for the click handler/active-tab comparison)
- `Schedule/index.tsx`: `key={index}` → `key={entry.hour}` (verified hours don't repeat within a single day's schedule in `edition/ScheduleDays.ts`)
- `CompanyHistorySection`'s skeleton loader: left the index key as-is with a comment explaining why — it's a fixed-count loading placeholder with no data yet, so there's no real identity to key on.

Note on scope: the original audit's "6 index-as-key sites" included `GiveawaySection`, which turned out to already be fixed by an unrelated prior refactor (`key={student.id}`) — so the actual count of live sites is 5, of which 4 get a real fix here and 1 (the skeleton loader) is a false positive, documented rather than changed.

The other half of the issue — a client component importing a type from a Prisma-backed module — is already resolved in the current codebase. No such import exists anymore, and `AGENTS.md` now explicitly documents type-only Prisma imports as acceptable practice.

### DIFF
--- a/src/components/Companies/CompanyProfile/CompanyHistorySection/index.tsx
+++ b/src/components/Companies/CompanyProfile/CompanyHistorySection/index.tsx
@@ -141,6 +141,8 @@ const CompanySavesSection = () => {
         style={{ scrollbarGutter: "stable" }}
       >
         {!historyData ? (
+          // fixed-count loading placeholders with no data yet, so there's no
+          // identity to key on besides position
           Array(3)
             .fill(1)
             .map((_, i) => (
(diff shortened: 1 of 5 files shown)


dinisjcorreia (approved):
Reviewed changes and caller data. Stable keys are unique within each rendered sibling list; fixed skeleton keys are appropriately documented. CI is green.

PR #199, opened 2026-07-14
refactor: move GET display flows to server components

### LINKED ISSUE #139: [Refactor] Move GET-and-display flows to Server Components
### Description
The App Router is used like a client-side SPA: data is fetched from the browser instead of rendered on the server.

### Why
107 of ~110 components are `"use client"`, and 23 fetch their own data via the API, causing extra bundle weight, waterfalls, and spinners.

### How
Do a boundary audit — push GET-and-display flows to Server Components (reusing lib/services); reserve `"use client"` for genuine interactivity; map DTOs at the server->client prop boundary.

### Dependencies
None (uses existing lib functions; the DTO mapping is tracked separately).

dinisjcorreia:
## Summary
- server-seed session data only inside the admin and profiles route groups; public routes keep static generation
- server-render interests, company history, and initial action QR data; pass DTOs into interactive client islands
- gate profile interest loading to the own-profile path and remove the unused client-side interest-matching flow

## Verification
- `pnpm test` (35 tests)
- `pnpm typecheck`
- `pnpm lint` (passes with 4 pre-existing formatting warnings)
- production build with placeholder public env values; `/`, `/about`, `/login`, privacy/reset pages are static, while admin/profiles and `/signup` are dynamic

## Runtime limitation
- DB-backed `/signup` could not be exercised successfully because this worktree has no local PostgreSQL/Supabase environment; failure was the expected Prisma connection error.


src/app/layout.tsx
```diff
@@ -12,17 +12,23 @@ import { AuthContextProvider } from "@/contexts/AuthContext";
...
+import { toSessionDto } from "@/application/dto/sessionDto";
+import getServerSession from "@/application/services/sessionService";
 import { branding } from "@/edition/branding";
 
 export const metadata: Metadata = {
   ...branding.metadata,
 };
 
-export default function RootLayout({
+export const dynamic = "force-dynamic";
```
david-fs-valente: **[Blocking] `force-dynamic` here removes static generation from every route in the app, not just the ones that need session data**

`RootLayout` now calls `getServerSession()` and this file sets `export const dynamic = "force-dynamic"`. Since this is the *root* layout, that setting cascades to every route in the tree.

I built this branch and `dev` locally to compare the Next.js route table directly:

| Route | `dev` (before) | this branch (after) |
|---|---|---|
| `/` | `○` Static | `ƒ` Dynamic |
| `/about` | `○` Static | `ƒ` Dynamic |
| `/login` | `○` Static | `ƒ` Dynamic |
| `/privacy-policy` | `○` Static | `ƒ` Dynamic |
| `/password-reset`, `/password-reset/confirm` | `○` Static | `ƒ` Dynamic |
| `/signup`, `/signup/employee` | `○` Static | `ƒ` Dynamic |

The homepage, About, Privacy Policy, and the auth entry pages all lose static prerendering and now hit `getServerSession()` (a Supabase call) on every request, even though none of those pages need per-request session data for anything other than seeding `AuthContextProvider`.

Suggest scoping the session lookup + `initialUser` seeding to a nested layout for the route groups that actually need it (`(profiles)`, `(admin)`, and wherever `(auth)` needs it), rather than the root layout — that keeps `/`, `/about`, `/privacy-policy`, etc. statically generated. If sitewide dynamic rendering is intentional, it's a big enough tradeoff (loses CDN caching + adds a DB round trip to every request including bots/crawlers hitting `/`) that it should be called out explicitly in the PR description rather than landing as a side effect.

david-fs-valente (changes requested):
Requesting changes on one blocking issue: forcing dynamic rendering on the root layout removes static generation for several public pages that don't need it. See inline comment.

(Left the minor/non-blocking finding and the merge-conflict/stacking notes as separate PR comments, not part of this review.)

david-fs-valente:
**[Minor, non-blocking] Unconditional `getInterests()` read in the student page**

In `src/app/(profiles)/student/[...data]/page.tsx`, `getInterests()` is now called unconditionally inside the `Promise.all`, but `availableInterests` is only consumed by the `ProfileSectionContainer` branch (student viewing their own profile). For the `CompanyViewProfileSectionContainer` and `PublicProfileSectionContainer` branches — arguably the more common views — it's a DB read that's fetched and then discarded.

It's parallelized so it doesn't add latency, but it does add unnecessary DB load on every request for two of the three render paths, which cuts a little against this PR's own "parallelize independent reads" performance goal. Not blocking — just worth gating the fetch on which branch will actually render, if you're touching this file again.

src/app/layout.tsx
dinisjcorreia: Fixed in 58105a3. Removed root session lookup and root force-dynamic. Session DTO seeding is now scoped to `(admin)` and `(profiles)` layouts; build verifies `/`, `/about`, `/login`, privacy/reset pages are static again. `/signup` remains explicitly dynamic because it server-loads interests and must not query Prisma during build.

dinisjcorreia:
Fixed in 58105a3. `getInterests()` now runs only for the own-profile branch, which is the only path that renders `ProfileSectionContainer` and consumes `availableInterests`.

david-fs-valente (changes requested):
## Review: refactor: move GET display flows to server components

Solid refactor overall — moving GET display flows to server components is a real win for TTFB and lets public routes stay statically generated. A few things should be addressed before merge.

### 1. (Medium) Company history no longer refreshes on client-side tab switches

`history` is now fetched once in `dashboard/page.tsx` and threaded down as a static prop through `CompanyProfileSectionContainer` → `CompanyStatsSection` / `CompanySavedProfilesSection` → `CompanyHistorySection`. Previously, `CompanySavesSection` fetched `/companies/history` itself on every mount, so switching tabs (`activeTab` is local state — `renderContent()`'s switch unmounts/remounts the tab content) always showed fresh data.

Now, `CompanySavedProfilesSection.handleScan` / `handleManualEntry` save a new student and just `router.push()` to the student's preview page — there's no `router.refresh()`. A company scanning several students in a row at a booth, then switching from "Scan de Perfil" back to "Sumário" (pure client state, no navigation), won't see newly-scanned students in the history list or count until a hard reload. (`CompanyStatsSection.handleSave()` — saving interests — does call `router.refresh()`, so that action incidentally re-syncs history, but the scan flow itself doesn't.)

Suggest adding `router.refresh()` after a successful save in `handleScan`/`handleManualEntry`, or re-fetching history client-side after a scan.

### 2. (Low) Orphaned interest-matching route left behind

The PR removes `InterestMatchingSection` (the only caller), but `src/app/api/interests/matching/[id]/route.ts` and `InterestMatchDto`/`toInterestMatchDto` in `src/application/dto/companyDto.ts` are left in place with no remaining caller anywhere in the codebase. Worth removing in this PR (or a fast follow) rather than leaving dead server code, since the PR description frames this as removing "the unused client-side interest-matching flow" — currently only half of that flow is removed.

### 3. (Low) Duplicated layout boilerplate

`(admin)/layout.tsx` and `(profiles)/layout.tsx` are identical (seed `AuthContextProvider` from `getServerSession()`, `export const dynamic = "force-dynamic"`). Not wrong, but a shared helper (e.g. a `SessionAuthLayout` wrapper) would avoid the two drifting independently later.

Requesting changes mainly on #1 since it's a visible regression to a core company-facing flow; #2 and #3 are fine to fold in here or defer to a follow-up.

david-fs-valente:
### Minor / non-blocking notes

- `toInterestDto`'s input type is `InterestDto` rather than the raw domain/repository type — but this matches the existing convention elsewhere in the codebase (`toCompanyDto`, `toInterestMatchDto` do the same), so it's consistent, not a new issue.
- The root `AuthContextProvider` in `src/app/layout.tsx` still performs its own client-side `fetchSession()` (since it passes no `initialUser`) on every route, including admin/profile pages, which now also get a server-seeded nested provider. That's pre-existing `Topbar` behavior, not introduced by this PR, but worth knowing it's a redundant network call on those routes specifically.

david-fs-valente:
A couple of non-blocking observations from review:

**Design note — duplicate `AuthContextProvider`:** The root `layout.tsx` still wraps the whole app in `<AuthContextProvider>` with no `initialUser`, which fires a client-side `fetchSession()` on mount (used by `Topbar`, which renders outside the route-group `children`). The new `(admin)`/`(profiles)` layouts add a *second*, nested `AuthContextProvider` via `SessionAuthLayout`, seeded from the server. That's a reasonable trade-off — it avoids forcing the whole app dynamic just to seed one context — but it does mean every admin/profile page load fires a redundant `/api/auth/session` client fetch (for `Topbar`'s copy of the context) in addition to the now-SSR'd session. Not a correctness issue since the inner provider correctly shadows the outer one for page content, just a small duplicated network cost worth being aware of if session-fetch volume ever becomes a concern.

**Minor:**
- `/api/interests/route.ts` is now unreferenced by any client code (its only two callers — `InterestSelector` and the deleted matching route — are gone). Fine to leave as a public API, but worth a quick check on whether it's still needed.
- `getActionQrCode()` in the admin actions page always signs a JWT even when `findActionById` returns null (checked after the fact) — pre-existing pattern carried over from the old route handler, not a regression, just a small amount of wasted work on the 404 path.

david-fs-valente (approved):
Reviewed the diff, ran tests/typecheck/lint, and spot-checked the touched flows (session, action QR, saved-student history, interests gating). Refactor is clean and consistent — data flows server fetch → DTO → client island prop throughout, dead interest-matching code is fully removed rather than left dangling, and new test coverage matches the server-seeded contract. Left a couple of non-blocking notes in a separate comment. LGTM.

dinisjcorreia:
Fixed non-blocking review notes in 3da97e6.

- Removed root AuthContext/Topbar; each route group now owns one provider + Topbar. Protected groups use server-seeded `SessionAuthLayout`, so no redundant `/api/auth/session` fetch.
- Removed unused `/api/interests` endpoint after confirming no callers.
- `getActionQrCode()` now returns before signing when action is missing; GET returns 404. Added regression test.

Verified: 35 tests, typecheck, build. Lint passes with four pre-existing formatting warnings.

PR #200, opened 2026-07-17
feat: add orphaned-file GC job

### LINKED ISSUE #136: [Feat] Add an orphaned-file GC job (storage <-> DB reconciliation)
### Description
Uploaded avatars/CVs are never cleaned up, so storage accumulates orphans.

### Why
`Student.avatar`/`Student.cv` are bare string IDs with no referential integrity.

### How
Use Supabase `pg_cron` to reconcile `storage.objects` against `Student.avatar`/`cv`, deleting only files that are unreferenced AND older than a grace window (e.g. 48h). Run daily.

### Dependencies
None (the schema-integrity work would reduce, not gate, this).

dinisjcorreia:
## Summary
- add pg_cron reconciliation for orphaned avatar and CV objects
- enforce 48-hour grace window, UUID/path scope, and 1,000-object cap
- delete through Supabase Storage API via pg_net using Vault credentials
- expose one shared candidate query for dry-run and deletion
- document cron/pg_net failure diagnostics

## Safety gate
1. Run supabase/storage-gc.sql manually. It installs functions but does not schedule deletion; its final query returns the exact candidates without deleting anything.
2. Review every returned bucket/path against Student.avatar and Student.cv.
3. Only after confirming the dry run, run supabase/storage-gc-enable.sql to schedule daily deletion.

SQL was not executed by this task because repository instructions prohibit agents from running SQL. Production activation therefore remains blocked on the required human dry run.

## Verification
- pnpm install --frozen-lockfile
- pnpm lint
- pnpm test (19 files, 51 tests)
- pnpm typecheck

Rebased onto current dev, including the merged React Toastify consolidation; no SweetAlert dependency or usage remains.

### DIFF
--- a/supabase/storage-gc.sql
+++ b/supabase/storage-gc.sql
@@ -1,0 +1,101 @@
+-- Run manually in the hosted Supabase SQL editor after creating these Vault
+-- secrets (Dashboard > Integrations > Vault):
+--   storage_gc_project_url      = https://<project-ref>.supabase.co
+--   storage_gc_service_role_key = the project's service-role key
+--
+-- storage.objects is metadata only. This job reads it for reconciliation, then
+-- deletes through the Storage API so both the object and its metadata disappear.
+
+create extension if not exists pg_net with schema extensions;
+create extension if not exists pg_cron;
+
+create or replace function public.queue_orphaned_storage_gc()
+returns integer
+language plpgsql
+security definer
+set search_path = ''
+as $$
+declare
+  candidate record;
+  project_url text;
+  service_role_key text;
+  request_headers jsonb;
+  queued_count integer := 0;
+begin
+  select decrypted_secret
+  into strict project_url
+  from vault.decrypted_secrets
+  where name = 'storage_gc_project_url';
+
+  select decrypted_secret
+  into strict service_role_key
+  from vault.decrypted_secrets
+  where name = 'storage_gc_service_role_key';
+
+  request_headers := jsonb_build_object(
+    'apikey', service_role_key,
+    'Authorization', 'Bearer ' || service_role_key
+  );
+
+  for candidate in
+    select objects.bucket_id, objects.name
+    from storage.objects as objects
+    where objects.created_at < now() - interval '48 hours'
+      and (
+        (
+          objects.bucket_id = 'avatars'
+          and objects.name ~ '^distribution/avatar/[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$'
+          and not exists (
+            select 1
+            from public."Student" as students
+            where students.avatar = replace(objects.name, 'distribution/avatar/', '')
+              or students.avatar = objects.name
+              or right(
+                split_part(students.avatar, '?', 1),
+                length('/avatars/' || objects.name)
+              ) = '/avatars/' || objects.name
+          )
+        )
+        or (
+          objects.bucket_id = 'cvs'
+          and objects.name ~ '^distribution/cv/[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\.pdf$'
+          and not exists (
+            select 1
+            from public."Student" as students
+            where students.cv = replace(
+                replace(objects.name, 'distribution/cv/', ''),
+                '.pdf',
+                ''
+              )
+              or students.cv = objects.name
+              or right(
+                split_part(students.cv, '?', 1),
+                length('/cvs/' || objects.name)
+              ) = '/cvs/' || objects.name
+          )
+        )
+      )
+    order by objects.created_at
+    -- ponytail: cap each run; raise it or run more often if backlog exceeds one day.
+    limit 1000
+  loop
+    perform net.http_delete(
+      url := rtrim(project_url, '/') || '/storage/v1/object/' ||
+        candidate.bucket_id || '/' || candidate.name,
+      headers := request_headers,
+      timeout_milliseconds := 5000
+    );
+    queued_count := queued_count + 1;
+  end loop;
+
+  return queued_count;
+end;
+$$;
+
+revoke all on function public.queue_orphaned_storage_gc() from public;
+
+select cron.schedule(
+  'storage-orphan-gc',
+  '0 3 * * *',
+  $job$select public.queue_orphaned_storage_gc();$job$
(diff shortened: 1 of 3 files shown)


supabase/storage-gc.sql
```diff
@@ -0,0 +1,101 @@
...
+                length('/cvs/' || objects.name)
+              ) = '/cvs/' || objects.name
+          )
+        )
+      )
+    order by objects.created_at
+    -- ponytail: cap each run; raise it or run more often if backlog exceeds one day.
+    limit 1000
+  loop
+    perform net.http_delete(
```
david-fs-valente, line 100: [Blocking] This deletion path (and the whole matching query above it) has never been executed against real data - the PR description confirms the SQL wasn't run. A wrong regex or reference-match here deletes a live student avatar/CV with no undo once the Storage API call succeeds. Please dry-run this (e.g. temporarily log/count candidates instead of calling net.http_delete, or test against a staging project) before this cron job goes live.

david-fs-valente (changes requested):
## Review

Adds a daily `pg_cron` job (`supabase/storage-gc.sql`) that deletes orphaned avatar/CV objects from Supabase Storage via `pg_net`, plus docs and an unrelated `sweetalert` dependency fix.

I cross-checked the "is this file still referenced" logic against the actual write paths (`studentService.ts`, `storage/avatar/route.ts`, `storage/cv/route.ts`) — `Student.cv` is a bare UUID and `Student.avatar` is a full public Storage URL, and the query's branches correctly match both real-world formats. The `SECURITY DEFINER`/`search_path = ''`/`revoke ... from public` hygiene and the `select ... into strict` secret lookups are all solid, and the script is safely idempotent (`create extension if not exists`, `cron.schedule` upserts by job name).

**Blocking:** the PR body says "SQL was not executed from this task." This function performs irreversible deletions against production student data based on regex/string matching that took cross-referencing three separate source files to confirm by hand — a subtle mismatch here has no rollback once the Storage API delete succeeds. Before this runs against real data, please dry-run it at least once: e.g. temporarily swap `perform net.http_delete(...)` for a `raise notice`/count of candidates (or point it at a staging project) and confirm the candidate set matches expectations before enabling the cron schedule for real.

Requesting changes on that basis — everything else I found is minor/non-blocking and posted separately.

david-fs-valente:
## Other findings (non-blocking)

- **Scope mixing:** the `sweetalert` dependency fix (`package.json`/`pnpm-lock.yaml`) is unrelated to the GC feature — it's a real pre-existing bug (`CompanyAdminSection` imports it but it was never declared as a dependency on `dev`), just worth splitting into its own commit/PR next time per repo convention. Low risk either way, it's purely additive.
- **No alerting on persistent delete failures:** `net.http_delete` is fire-and-forget via `pg_net`, so a permanently broken delete (e.g. a rotated/expired service-role key) just silently retries forever via next-day rescans, with no visibility beyond manually checking `cron.job_run_details`/`net._http_response`. Worth a note on where to look if the backlog doesn't shrink.
- **Nit:** `-- ponytail: cap each run; raise it or run more often if backlog exceeds one day.` (in `supabase/storage-gc.sql`, near the `limit 1000`) — "ponytail" looks like a stray/leftover word, probably meant to be `-- note:`.
- **FYI:** `pnpm-lock.yaml` also bumps `postcss` 8.5.6 → 8.5.19 as an incidental transitive resolution from adding `sweetalert`. Almost certainly harmless (same minor range), just flagging in case anything Tailwind/PostCSS-related looks different in a build.

supabase/storage-gc.sql
dinisjcorreia: Addressed in 55a6060. The installer now exposes one shared storage_gc_candidates() query and runs it as a non-destructive final dry run. It no longer schedules deletion. A separate storage-gc-enable.sql must be run manually only after every candidate is reviewed. Repository instructions prohibit agents from running SQL, so production activation remains explicitly blocked on that human dry run.

dinisjcorreia:
Addressed all review findings:

- Blocking dry-run: installer no longer enables cron; it ends with the shared non-destructive candidate query. Separate enable SQL requires manual execution after review.
- Scope: SweetAlert removed from this PR and split into #201.
- Failure visibility: README now includes cron.job_run_details, net._http_response, and persistent-backlog queries plus pg_net response-retention note.
- Nit: changed the internal ponytail comment to note.
- Incidental PostCSS bump: removed; #201 preserves PostCSS 8.5.6.

I did not execute SQL because AGENTS.md explicitly prohibits agents from doing so. The deployment flow now prevents activation until a human performs and confirms the dry run.

dinisjcorreia:
Fresh CI confirms the only failure is the pre-existing undeclared SweetAlert import (`TS2307`), now isolated in #201. #201 CI is green. After #201 merges into `dev`, rerun/rebase this PR; GC lint/tests are green and no GC-specific CI failure exists.

dinisjcorreia:
Correction to prior CI note: #201 now removes the remaining SweetAlert usage by migrating CompanyAdminSection to React Toastify; it does not add SweetAlert. GC PR remains clean and depends on that independent baseline fix for typecheck.

dinisjcorreia:
Rebased onto latest dev after #201 merged. Fresh local verification passes: frozen install, lint, 19 test files / 51 tests, and typecheck. PR diff remains scoped to README plus the two storage GC SQL files. Fresh CI is running.

david-fs-valente (approved):
## Approved

All the previous review findings were addressed: the installer is now split into a non-destructive dry run (`storage-gc.sql`) and a separate, explicit enable step (`storage-gc-enable.sql`), the unrelated SweetAlert fix was split out into #201, and README now documents failure/backlog monitoring. Code and docs look correct and safe to ship.

**Holding the merge**, though, until the operational side is actually done:

- The dry run (`select * from public.storage_gc_candidates();`) needs to be run against the real hosted database and its output reviewed against `Student.avatar`/`Student.cv` by a human, per the PR's own safety gate.
- Only after that review confirms the candidate set is sane should `storage-gc-enable.sql` be run to schedule the cron job.
- Worth also revisiting whether this gap should be closed properly with an `Upload`-tracking table (referenced in the `schema.prisma` comment on `Student.avatar`/`cv`) rather than relying on this reconciliation job long-term — if that's planned as follow-up work, let's make sure it's tracked, and this GC job doesn't end up being the permanent fix for what's a missing DB-level integrity constraint.

I'll merge this once the dry run has been executed and reviewed, the cron job is confirmed active, and we've confirmed whether any further associated work (e.g. an uploads table) is in scope before or after this lands.

PR #201, opened 2026-07-17
fix: use React Toastify for company notifications

dinisjcorreia:
## Summary
- replace the remaining SweetAlert import in CompanyAdminSection with React Toastify
- preserve existing success messages for company create/update
- keep the repository on its single notification library

No package.json or pnpm-lock.yaml changes.

## Verification
- pnpm install --frozen-lockfile (removed local SweetAlert packages)
- pnpm lint
- pnpm test (17 files, 41 tests)
- pnpm typecheck

Split from #200 to keep the storage GC feature scoped.

### DIFF
--- a/src/components/CompanyAdminSection/index.tsx
+++ b/src/components/CompanyAdminSection/index.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import React, { useRef, useState } from "react";
-import Swal from "sweetalert";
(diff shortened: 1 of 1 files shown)


dinisjcorreia:
Corrected after notification-library clarification: dependency addition removed entirely. This PR now changes only CompanyAdminSection, replacing both SweetAlert calls with the established React Toastify API.

Issue #207, opened 2026-07-19
[Chore] Run a full security audit before the 2026 edition cutover


david-fs-valente:
### Description
The codebase hasn't had a fresh, full security pass. Several issues already found (e.g. the QR anti-replay check being disabled, compounded by a seconds/ms unit bug in token expiry) came out of ad hoc review, not a systematic audit — other issues may still be sitting undiscovered.

### Why
This app handles student PII (CVs, contact info) and company data, and the 2026 edition's `1.0.0` cutover is explicitly gated on the current backlog of architecture/security/correctness issues being resolved (see AGENTS.md). A full pass now is cheaper than finding a gap after the event is live and data is flowing.

### How
Audit pass covering: auth flows (session + hand-rolled JWT token paths), storage/upload paths (avatar/CV, public vs. private bucket boundaries), token handling (QR action tokens, CV export tokens, password reset), and CORS/deploy config. Produce a findings list, triage by severity, and open follow-up issues per finding rather than fixing inline during the audit.

### Dependencies
None directly, but test-coverage expansion is explicitly scoped to run after this audit surfaces which paths matter most — sequence this first.

david-fs-valente:
Audit complete. Covered auth flows (session + JWT token paths), storage/upload paths, token handling, and CORS/deploy config, cross-checked against issues already open before this pass (#212, #213, #211, #208) and the gotchas already documented in AGENTS.md (missing CSP, SavedStudent grain mismatch, dead bcrypt code) to avoid duplicating known findings.

**Findings, triaged by severity, filed as separate follow-ups per the process above:**

- **High** — #215: Avatar/CV upload endpoints have no authentication check at all (distinct from #213's rate-limiting gap; the in-code comment blaming this on #26 is stale — #26 is closed and unrelated).
- **Low** — #216: Missing ownership check on `/api/students/[code]/stats`.
- **Low** — #217: Student codes generated with `Math.random()`, not a CSPRNG.
- **Low** — #218: Docker `migrate` stage runs as root, unlike the `runner` stage.
- **Low** — #219: Sentry breadcrumb `message` field bypasses PII redaction (no live leak found, but unguarded).
- **Low/code-quality** — #220: `.match()` used instead of `===` in a student profile-owner check (landmine, not currently exploitable).
- **Chore** — #221: `innerHTML` cleanup in `ImportCvSection` (informational, no real security impact).
[...]

PR #228, opened 2026-07-20
fix: generate student codes with a CSPRNG

### LINKED ISSUE #217: [Fix] Student codes generated with Math.random(), not a CSPRNG
### Description
`src/utils/GenerateCode.ts` builds the 4-character public student code (`A-Z0-9`, ~1.68M keyspace) using `Math.random()`. Surfaced by the #207 security audit.

### Why
These codes function as bearer-ish identifiers in profile URLs and the giveaway/scan flow. `Math.random()` is not cryptographically secure and, combined with the small keyspace, makes brute-force enumeration of valid codes easier than it should be. Not independently critical on its own — the profile/CV/history endpoints all layer proper ownership checks on top — but it's the weak link that makes gaps like the missing ownership check on `/api/students/[code]/stats` exploitable via brute-force rather than just insider guessing.

### How
Generate codes using a CSPRNG (e.g. Node's `crypto.randomInt`/`crypto.randomBytes`) instead of `Math.random()`.

### Dependencies
None.

david-fs-valente:
Replaces `Math.random()` with `crypto.randomInt()` in `GenerateCode.ts` so the 4-character student codes aren't predictable.

Refs #217, part of #207.


dinisjcorreia (approved):
No findings. CSPRNG replacement preserves code shape and existing collision handling.

PR #229, opened 2026-07-20
fix: redact Sentry breadcrumb message field

### LINKED ISSUE #219: [Fix] Sentry breadcrumb message field bypasses PII redaction
### Description
`sanitizeSentryEvent` (`src/lib/sentryPrivacy.ts`) scrubs `breadcrumb.data` by key-name matching but never touches `breadcrumb.message`. Sentry's default browser `Breadcrumbs` integration captures `console.error(...)` calls as breadcrumbs whose `message` is the stringified error/args — that text would bypass redaction entirely if it ever contained PII. Surfaced by the #207 security audit.

### Why
Not currently exploitable: all interpolated `throw new Error(...)` messages in the codebase were checked and none embed real PII values (email, token, code) today — only safe fields like domain/year label/tier string. But it's a latent gap: nothing stops a future `console.error`/thrown error from including PII and leaking it into Sentry via this unsanitized field, and the app handles student PII (CVs, contact info) end to end.

### How
Extend `sanitizeSentryEvent` to also scrub or redact `breadcrumb.message` (same key/value heuristics used for `breadcrumb.data`), or disable console breadcrumbs if they're not needed for debugging.

### Dependencies
None.

david-fs-valente:
`sanitizeSentryEvent` redacted breadcrumb `data` by key but left `breadcrumb.message` untouched, so console-error-derived breadcrumb text could bypass PII redaction. Now blanked to the same placeholder used for the top-level event message.

Refs #219, part of #207.


dinisjcorreia (approved):
Reviewed Sentry breadcrumb-message redaction and coverage. No blocking issues found.

PR #230, opened 2026-07-20
fix: enforce ownership check on student stats route

### LINKED ISSUE #216: [Fix] Missing ownership check on /api/students/[code]/stats
### Description
`src/app/api/students/[code]/stats/route.ts` uses `auth: "session"` with no `authorize` callback, unlike every sibling student route (`[code]/route.ts`, `[code]/cv/route.ts`, `[code]/history/route.ts`), which all gate access via `isAllowedToViewStudent` or an explicit `authorize`. Surfaced by the #207 security audit.

### Why
Any logged-in student or employee can query `{totalScans, totalSaves}` for any other student by supplying their 4-character code, regardless of whether they're allowed to view that student's profile. Impact is minor on its own (aggregate counts only, no PII), but it's inconsistent with the access-control pattern used everywhere else for this exact resource, and is compounded by the weak code generation in a related finding (student codes aren't a CSPRNG, making enumeration easier).

### How
Add the same ownership/visibility check used by the sibling routes (`isAllowedToViewStudent` or an equivalent `authorize` callback) to this route.

### Dependencies
None.

david-fs-valente:
`/api/students/[code]/stats` only checked `auth: "session"`, unlike sibling student routes that enforce `isAllowedToViewStudent`. Any logged-in student/employee could query any other student's stats by code. `getStudentStats` now takes an access context and enforces the same ownership rule as `getStudentProfile`/`getStudentCv`.

Refs #216, part of #207.


src/app/(profiles)/student/[...data]/page.tsx
```diff
@@ -77,7 +77,11 @@ const StudentPage = async (props: ProfileProps) => {
 
   const [globalStats, todayStats, companies, history, actions, interests] =
     await Promise.all([
-      getStudentStats(student.code),
+      getStudentStats(student.code, {
```
Comment (dinisjcorreia) [inline src/app/(profiles)/student/[...data]/page.tsx]: **[P1] Preserve unsaved employee QR previews**\n\nA company can open a valid `preview` JWT before saving the student (`isPreview` bypasses the saved-profile check above), but this call now passes a company access context to `getStudentStats`. `isAllowedToViewStudent` then rejects it because `isStudentSaved` is false, so this `Promise.all` rejects and the whole preview page errors. The preview component does not use `globalStats` or `todayStats`; skip these stats calls for previews, or explicitly model the verified preview token as authorized access.

dinisjcorreia (changes requested):
Blocking: valid employee QR-preview sessions for students the company has not saved now fail before the preview renders.

src/app/(profiles)/student/[...data]/page.tsx
Comment (david-fs-valente) [reply src/app/(profiles)/student/[...data]/page.tsx]: Addressed in e7f496c: skip both `getStudentStats`/`getTodayStudentStats` entirely when `isPreview`, since `PreviewProfileSectionContainer` never reads either value. Ownership check itself is untouched for all real (non-preview) access paths.

david-fs-valente:
## Non-blocking suggestions

- **Redundant DB round-trip for the employee/company view path.** In `src/app/(profiles)/student/[...data]/page.tsx`, the page already computes `isSavedStudent` via `isSaved(session.employee.company.id, student.code)` and returns 404 earlier if `session.employee && !isSavedStudent && !isPreview`. By the time the `Promise.all` runs `getStudentStats(...)` for a non-preview employee view, `isAllowedToViewStudent` re-queries `isStudentSaved` a second time to reach the same true result. Not a correctness issue (CI is green, behavior is correct), just an extra Prisma round-trip per profile-view request that could be avoided by passing the already-known `isSavedStudent` result through instead of re-deriving it inside `getStudentStats`.
- **Test 2 in `savedStudentService.test.ts`** ("a student cannot query an unrelated student's stats") sets up `vi.mocked(isStudentSaved).mockResolvedValue(false)`, but since the `access` passed has no `companyId`, `isAllowedToViewStudent` short-circuits on `access.companyId` being falsy and never calls `isSaved` — the mock is dead setup in this test. Harmless, but could be dropped (or the test could pass a `companyId` to actually exercise that branch) for clarity.

david-fs-valente:
Re-reviewed the current diff (HEAD `2007718`) from scratch.

Both points from the prior non-blocking-suggestions comment are now resolved:
- The redundant `isStudentSaved` round-trip in the page render path is gone — `e94fd6f` threads the already-computed `isSavedStudent` through `StudentAccess.isSaved`, and `isAllowedToViewStudent` short-circuits on it (`access.isSaved ?? (await lookupIsSaved(...))`) instead of re-querying.
- The dead-mock test is fixed — `2007718` adds `companyId: "company-1"` to the "unrelated student" case and asserts `isStudentSaved` was actually called, so it now exercises the branch it claims to.

Also traced @dinisjcorreia's earlier P1 `CHANGES_REQUESTED` review (unsaved employee QR previews failing before render) — that was already fixed by `e7f496c` (stats calls skipped entirely for `isPreview`), independent of today's two commits, and confirmed still correct in the current code. Only calling this out because GitHub's `reviewDecision` is still showing `CHANGES_REQUESTED` — looks like it just needs a re-review/dismissal from @dinisjcorreia to clear, since the underlying code issue is resolved.

No new blocking or non-blocking issues found in this pass. CI is green.

PR #231, opened 2026-07-20
fix: use strict equality for student profile owner check

### LINKED ISSUE #220: [Fix] Replace .match() with strict equality in student profile-owner check
### Description
`src/app/(profiles)/student/[...data]/page.tsx:63` uses `!session.student.code.match(student.code)` — `String.prototype.match` with the target's code used as a live regex pattern — instead of strict equality (`!==`), to check whether the logged-in student owns the profile being viewed. A different branch of the same file (line 131) does the equivalent check with strict `===`. Surfaced by the #207 security audit.

### Why
Not currently exploitable: both operands are always fixed 4-character strings from a restricted charset (`[A-Z0-9]{4}`, see the related student-code-generation issue), so a substring-match of equal length is mathematically equivalent to equality today, and no regex metacharacters are reachable. It's flagged as a landmine, not a live bug — if code length or charset ever changes, this silently becomes exploitable (a regex-metacharacter-containing code could match more broadly than intended), and it's inconsistent with the strict equality used elsewhere in the same file for the same conceptual check.

### How
Replace `.match()` with strict equality (`!==`), consistent with the check at line 131.

### Dependencies
None.

david-fs-valente:
Replaces a `.match()`-as-regex ownership check with strict equality, consistent with the equivalent check elsewhere in the same file. Not currently exploitable given the fixed 4-char code charset, but a landmine if that ever changes.

Refs #220, part of #207.


dinisjcorreia (approved):
Reviewed owner-code comparison fix. `===` correctly performs literal identity comparison and matches existing owner check below. No blocking issues found.

Validated: `pnpm lint`, `pnpm test` (51 passing), `pnpm typecheck`.

PR #232, opened 2026-07-20
chore: replace innerHTML with React state in ImportCvSection

### LINKED ISSUE #221: [Chore] Replace innerHTML with React state in ImportCvSection
### Description
`src/components/Profile/ImportCvSection/index.tsx:33-37` sets a label's `innerHTML` directly from the selected file's own `.name`, bypassing React state/JSX rendering. Surfaced by the #207 security audit.

### Why
Informational only, not a real vulnerability: it's the user's own local file selection, rendered only in their own browser tab, never persisted or shown to anyone else — no exploitable impact. Flagged for idiomatic-React cleanup rather than security value, since manual DOM manipulation like this is easy to accidentally extend into a real gap later.

### How
Replace the direct `innerHTML` assignment with React state (store the selected filename in state and render it via JSX).

### Dependencies
None.

david-fs-valente:
Replaces a direct `innerHTML` assignment with React state for displaying the selected CV filename. Informational-only finding (own local file selection, never persisted/shown elsewhere) flagged for idiomatic-React cleanup.

Refs #221, part of #207.


dinisjcorreia (approved):
Reviewed changes. React state correctly replaces direct DOM mutation, selected-file behavior is covered, and no regressions found.

PR #233, opened 2026-07-20
fix: require session for avatar/CV uploads

### LINKED ISSUE #215: [Fix] Avatar and CV upload endpoints have no authentication check
### Description
`src/app/api/storage/avatar/route.ts` and `src/app/api/storage/cv/route.ts` bypass `defineHandler` entirely and have no session check of any kind. Both routes carry a code comment claiming the missing auth is "tracked separately (#26)" — but #26 is a closed, unrelated 2025-edition task ("Update Prisma Seed"), not a live tracker for this gap. This is functionally an untracked, currently-exploitable finding surfaced by the #207 security audit.

Content-type/magic-byte validation and size limits do exist on both routes, so this isn't unrestricted-file-type upload — but no account or session is required to reach them at all.

### Why
An anonymous, unauthenticated visitor can POST files directly into either Supabase Storage bucket:
- `avatars` is a **public** bucket, so every accepted upload immediately gets a public URL under the app's own trusted domain — usable for anonymous image hosting, defacement, or phishing content with zero attribution.
- `cvs` is a **private** bucket meant to hold PII-adjacent documents, but nothing stops mass anonymous uploads filling it with junk files — a storage-cost/quota abuse vector independent of anything CV-content-related.

This is distinct from #213 (rate limiting on uploads): #213 only throttles request *volume* for already-authenticated abuse; it doesn't address that these endpoints require no authentication at all. Rate limiting alone would still leave a single anonymous malicious upload possible.

### How
Wire both routes through `defineHandler` (or add an equivalent manual session check, since these stay as plain route exports per AGENTS.md's multipart/form-data exception) requiring at minimum `auth: "session"`, with ownership scoping so a user can only write their own avatar/CV. Update or remove the stale `#26` comment reference in both files.

### Dependencies
None — can be fixed independently of #213 (rate limiting is a separate, complementary control).

david-fs-valente:
`/api/storage/avatar` and `/api/storage/cv` had no auth check at all, so anyone could upload directly into Supabase Storage. Adding a session check outright would have broken signup, since `FinalStep` currently uploads avatar/CV *before* the account exists. Reordered the signup flow to create the Supabase Auth account first, then upload (now authenticated), then create the Student profile — and gated both upload routes on `auth: "session"`.

Also fixed two now-dead-but-still-typechecked signup step components (`AvatarStep`/`CvStep`, already marked deprecated and unused) so `pnpm typecheck` stays green — no behavior/wiring change to either.

Refs #215, part of #207.


dinisjcorreia (changes requested):
[P1] `src/components/SignUp/FinalStep/index.tsx:117` — preserve a retry path after account creation. `createAccount()` has already created the Supabase user when either upload below or `createStudentProfile()` fails, but every retry begins with `/auth/signup` again. That returns a duplicate-account error, so the student cannot finish creating their profile and remains unable to use the app. Either defer account creation until all fallible pre-work succeeds, or make this flow resume profile creation for authenticated existing users (with cleanup/rollback for failed uploads).

david-fs-valente:
Addressed in 3cec697: check for an existing session before calling \`createAccount()\`. If one exists with no student profile yet (account created but a later step failed, or the user came back later), resume straight into upload + profile creation instead of retrying account creation. If a profile already exists too, just redirect home. Covers both the same-session retry and the return-later case.

david-fs-valente:
## Non-blocking suggestions

- **Cross-role session edge case in `FinalStep.handleSubmit`:** the new resume logic checks `existingSession?.student` to detect an already-created profile, and only calls `createAccount` when `!existingSession`. If a user is already authenticated as a *non-student* (e.g. an employee/admin session in the same browser) and lands on the public student signup flow, `existingSession` is truthy but has no `.student`, so the code skips account creation entirely and goes straight to avatar/CV upload + `POST /students` using that unrelated session. The backend's `authorize: (session) => session.role === "STUDENT"` on `/api/students` correctly rejects this (no security issue), but the user gets an unhelpful generic error ("Ocorreu um erro ao criar o perfil.") instead of a clear "you're already logged in as something else" message. Probably rare in practice, but worth a guard/clearer message if it's easy.

- **Inconsistent upload-failure handling:** `handleAvatarUpload` swallows upload failures (toast + continue with `avatarUrl = null`), while the newly-added CV upload path hard-stops the whole submission on failure (`toast.error(...); return setLoading(false);`). The avatar behavior predates this PR, but now that CV upload sits right next to it with different semantics, the inconsistency is more visible — worth a follow-up for consistent behavior (either both block or both degrade gracefully).

- **Test coverage:** the new `route.test.ts` files for `storage/avatar` and `storage/cv` only cover the 401-unauthenticated case (the actual behavior change), which is the right priority. No test exists for the more complex client-side resume logic in `FinalStep.handleSubmit` (existing session / existing student / fresh signup branches) — understandable since `SignUp/*` has no existing test coverage at all, but it's the riskiest new logic in this PR and would benefit most from a test if there's appetite.

david-fs-valente:
Re-reviewed the current diff (as of ea86b1d/44f81f3/50cfd31) — all three previously-flagged non-blocking suggestions are now addressed:

1. **Cross-role session edge case** — fixed in `ea86b1d`: `StudentSignUp` now redirects non-student sessions away on mount, and `FinalStep.handleSubmit` has a matching defense-in-depth check with a clear toast ("Já tens sessão iniciada como outro tipo de conta.") instead of falling through to a generic error.
2. **Inconsistent upload-failure handling** — fixed in `44f81f3`: CV upload failure now toasts and continues (`cvId` left `undefined`) instead of hard-stopping the submission, matching the avatar behavior. `postStudentSchema`'s `cvId`/`avatarUrl` are both optional, so this is safe server-side too.
3. **Test coverage** — `50cfd31` adds `FinalStep/index.test.tsx` covering all four resume branches (fresh signup, resume-with-no-profile, profile-already-exists redirect, non-student-session block).

Re-analyzed the full current diff from scratch (not just these three points) — no new blocking or non-blocking issues found. CI (Test, Typecheck & Lint) is green.

One process note unrelated to the code: dinisjcorreia's earlier `CHANGES_REQUESTED` review (on the duplicate-account retry issue, addressed same-day in `3cec697`) is still showing as the latest formal review decision — worth a re-review/dismissal for a clean merge state, since GitHub won't surface that it's already resolved.

PR #234, opened 2026-07-20
fix: run Docker migrate stage as non-root

### LINKED ISSUE #218: [Fix] Docker migrate stage runs as root, unlike the runner stage
### Description
The `runner` stage in the `Dockerfile` explicitly drops to a non-root user (`USER nextjs`), but the `migrator` stage (used by `docker-compose.app.yml`'s `migrate` service) inherits directly from `builder`, which never sets `USER` — so `npx prisma migrate deploy` runs as root inside its container. Surfaced by the #207 security audit.

### Why
Low practical risk on its own — it's a short-lived, non-network-facing job that only runs `prisma migrate deploy` against `DATABASE_URL` — but it's a defense-in-depth gap: a vulnerability in the Prisma CLI or a malicious migration file would execute as root rather than a restricted user, unlike every other container in the stack.

### How
Add a `USER` step to the `migrator` stage (reuse or recreate the same non-root user used in `runner`) before the `CMD`/entrypoint.

### Dependencies
None.

david-fs-valente:
The `migrator` stage ran `prisma migrate deploy` as root, unlike `runner`. Moved the non-root user creation into the shared `base` stage and added `USER nextjs` to `migrator`. Verified with `docker build --target migrator/runner` + `docker run --entrypoint whoami`.

Refs #218, part of #207.


david-fs-valente:
Added one more commit here (6ca1fdd) that's not strictly in scope of #218 — an environment-related fix found while verifying the Docker build: `NEXT_PUBLIC_BASE_URL`'s empty-string Dockerfile ARG default was bypassing its existing Zod `.default()`, causing `next build` to fail when no build args are supplied. Same class of fix already applied to `NEXT_PUBLIC_SENTRY_DSN`. Verified with a real `docker build` before/after.

dinisjcorreia (approved):
Reviewed changes. No blocking findings. Focused env test and lint pass locally; CI is green.

PR #236, opened 2026-07-20
fix: pass real seconds to QR action token expiresIn

### LINKED ISSUE #212: [Fix] Fix QR anti-replay window (currently ~8.3 hours instead of 30s)
### Description
QR action-scan tokens are valid for ~8.3 hours instead of the intended 30 seconds: `getActionQrCode()` passes a millisecond value (`actionQrCodeRefreshRateMs * 2`) directly into `jsonwebtoken`'s `expiresIn`, which is interpreted as seconds. Since QR scans earn points toward the giveaway, a screenshotted QR code can currently be replayed for hours after it's displayed.

### Why
This directly affects a real prize (the giveaway) and is trivially exploitable — screenshot a QR code, redeem it anytime within the window. The fix is small and there's no rational reason to leave it as-is.

### How
Fix the units bug so `expiresIn` receives real seconds. `jsonwebtoken`'s own `exp` check (already active by default in `verifyJwt`, since nothing passes `ignoreExpiration`) then enforces the 30-second freshness window with no extra code, DB reads, or caching layer needed. Smoke-test manually post-deploy — one edge case (e.g. an unusual restriction on the production anon key) can't be fully ruled out from static analysis alone.

### Dependencies
None.

david-fs-valente:
`getActionQrCode()` passed `actionQrCodeRefreshRateMs * 2` (a millisecond value) straight into `jsonwebtoken`'s numeric `expiresIn`, which is interpreted as seconds — making the QR action token valid for ~8.3 hours instead of 30 seconds. Since scans earn giveaway points, a screenshotted code was replayable for hours. Fixed by dividing by 1000; `jsonwebtoken`'s own `exp` check (already active in `verifyJwt`) now enforces the 30s window with no extra code needed.

Refs #212.

### DIFF
--- a/src/application/services/actionService.test.ts
+++ b/src/application/services/actionService.test.ts
@@ -32,3 +32,16 @@ test("does not sign a QR token for a missing action", async () => {
   await expect(getActionQrCode("missing")).resolves.toBeNull();
   expect(signJwt).not.toHaveBeenCalled();
 });
+
+test("signs the QR token with a real ~30s expiry, not 30000 seconds", async () => {
+  vi.mocked(findActionById).mockResolvedValue({
+    id: "action-1",
+  } as Awaited<ReturnType<typeof findActionById>>);
+
+  await getActionQrCode("action-1");
+
+  expect(signJwt).toHaveBeenCalledWith(
+    expect.objectContaining({ id: "action-1" }),
+    expect.objectContaining({ expiresIn: 30 })
+  );
+});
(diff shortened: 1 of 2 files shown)


dinisjcorreia (approved):
Reviewed expiry conversion and QR scan flow. Numeric `expiresIn: 30` now matches jsonwebtoken semantics; regression coverage added. Tests, lint, and typecheck pass.

PR #237, opened 2026-07-20
fix: stop migrate service from running a full Next.js build

### LINKED ISSUE #206: [Fix] Stop migrate service from running a full Next.js build
### Description
The Dockerfile builds `migrate` `FROM builder AS migrator`, so the `migrate` service goes through the entire `pnpm run build` (full Next.js/webpack compile) just to get the Prisma CLI, generated client, `schema.prisma`, and `migrations/`. Docker Compose builds `web` and `migrate` concurrently by default (BuildKit parallelizes independent targets), so for a chunk of every deploy there are two full Next.js production builds running at once — likely the single biggest contributor to a build-time memory-pressure incident on 2026-07-18. `NODE_OPTIONS="--max-old-space-size=2048"` is already set per build process, so two concurrent builds can approach ~4GB+ before anything else is counted.

### Why
We're a non-profit running the app, DB, and Coolify itself on one VPS — a recurring memory-pressure incident is a cost problem as much as a reliability one: the two ways out are "stop doing the wasteful thing" (free) or "provision more RAM" (recurring monthly cost). This fixes the actual cause for free.

### How
Restructure the Dockerfile so `migrate` branches off the `builder` stage **before** `pnpm run build` (right after `prisma generate`), instead of reusing the full post-build `builder` stage. `migrate` never touches `.next/` or webpack output, so this needs no functional trade-off — should cut build-time memory pressure roughly in half (one real Next.js build instead of two) and speed up deploys.

### Dependencies
None directly, but touches the same Dockerfile stages as the pnpm cache-mount issue (#205) — worth sequencing together or in the same PR.

david-fs-valente:
`migrate` reused the full post-build `builder` stage just to get the Prisma CLI/schema/migrations, so every deploy ran two full Next.js builds concurrently (web + migrate) — a real build-time memory-pressure incident. Split `builder` into a lean prisma-only stage (`migrator` branches off here, before the Next.js build) and `app-builder` (does the actual `next build`). Verified both `--target migrator` (no webpack step now) and `--target runner` (still builds correctly) with real `docker build` runs.

Note: this and #234 (#218, already open) both touch the `migrator`/`builder` stage boundary — whichever merges second will need a small conflict resolution.

Refs #206.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,26 @@ jobs:
 
       - name: Lint
         run: pnpm lint
+
+  docker-migrator:
+    name: Docker migrator stage runs as non-root
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      # Only builds the prisma-only `migrator` stage (no Next.js build), so
+      # this stays cheap. Building alone wouldn't catch a stage silently
+      # losing its USER instruction - `docker build` succeeds either way -
+      # so also run the image and assert the user, which is the actual
+      # regression this guards against (see Dockerfile's migrator stage).
+      - name: Build migrator stage
+        run: docker build --target migrator -t migrator-check .
+
+      - name: Verify migrator runs as a non-root user
+        run: |
+          user="$(docker run --rm --entrypoint whoami migrator-check)"
+          if [ "$user" != "nextjs" ]; then
+            echo "Expected migrator stage to run as 'nextjs', got '$user'" >&2
+            exit 1
+          fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,24 @@ FROM base AS builder
 WORKDIR /app
 COPY --from=deps /app/node_modules ./node_modules
 COPY --from=deps /app/package.json /app/pnpm-lock.yaml ./
+COPY prisma ./prisma
 
+# Only needs node_modules + the schema/migrations, not the rest of the app
+# source - deliberately kept minimal so `migrator` (which branches off here)
+# never pays for the Next.js build in `app-builder` below.
+RUN npx prisma generate
+
+# Standalone stage to apply pending migrations against DATABASE_URL before
+# the app starts. Branches off `builder` *before* the Next.js build below,
+# so `migrate` only pays for `prisma generate`, not a full webpack compile -
+# Docker Compose builds `web`/`migrate` concurrently, and two full builds at
+# once was a real build-time memory-pressure incident. See
+# docker-compose.app.yml's `migrate` service.
+FROM builder AS migrator
+USER nextjs
+CMD ["npx", "prisma", "migrate", "deploy"]
+
+FROM builder AS app-builder
 ARG NEXT_PUBLIC_SENTRY_DSN=""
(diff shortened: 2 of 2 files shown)


david-fs-valente:
## Blocking issues

**Rebase/conflict resolution will silently drop the non-root `migrate` fix.**

This branch's merge-base (`52e3552`) predates `dev`'s commit `751cc7e` ("fix: run Docker migrate stage as non-root"), which made two coordinated changes that this PR's new stage layout doesn't carry forward:

1. It moved the `addgroup`/`adduser` (creating the `nodejs`/`nextjs` user) up into the `base` stage specifically so it's available before `migrator` branches off — this PR's `base` stage still creates that user only down in `runner`.
2. It added `USER nextjs` before the `migrator` stage's `CMD` — this PR's new `migrator` stage (`FROM builder AS migrator`) has no `USER nextjs` line at all.

`gh pr view` already reports `mergeable: CONFLICTING` for this exact reason (both this PR and `dev` touch the `builder`/`migrator` boundary). The PR body flags a conflict with #234, but not this one. If the conflict is resolved by mechanically taking this branch's version of the Dockerfile, `migrate` goes back to running as root — silently undoing a recent, deliberate hardening fix, and CI won't catch it since nothing here builds/inspects the Docker image.

**Fix before merge:** rebase onto current `dev`, keep the `nodejs`/`nextjs` user creation in the `base` stage, and add `USER nextjs` before the `CMD` in the new (earlier-branching) `migrator` stage.

david-fs-valente:
## Non-blocking suggestions

- **Pre-existing redundancy in `app-builder`:** `COPY package.json package-lock.json* pnpm-lock.yaml* ./` re-copies files already present in `app-builder` via inheritance from `builder` (which copies `package.json`/`pnpm-lock.yaml` from the `deps` stage a few lines up). Not introduced by this PR and harmless, but since you're already touching this stage boundary it'd be a free cleanup.
- **No CI signal for Docker build regressions:** this class of bug (a stage losing `USER nextjs`) isn't caught by lint/typecheck/test — only by a real `docker build`, which the PR description says was done manually. A lightweight CI step that builds the `migrator` target (even just `docker build --target migrator .`) would catch this kind of regression automatically going forward. Optional, out of scope for this PR.

david-fs-valente:
## Re-review: previously-flagged issues resolved

Verified directly against the current PR head and current `dev` tip, not just the `mergeable` field:

- **Blocking issue (root `migrate` container) — fixed.** The branch's merge-base with `dev` is now `dev`'s exact tip (`7829df6`), so `751cc7e` is fully incorporated. `base` still creates the `nodejs`/`nextjs` user (inherited, untouched by this PR), and the new `migrator` stage (`FROM builder AS migrator`) now has `USER nextjs` before its `CMD`. `mergeable` is `MERGEABLE`, no conflict remains on this stage boundary.
- **Non-blocking: redundant `COPY package.json ...` in `app-builder`** — removed, with a comment explaining it's already inherited from `builder`.
- **Non-blocking: no CI signal for Docker regressions** — addressed with a new `docker-migrator` CI job that builds `--target migrator` and asserts the container runs as `nextjs` via `whoami`, not just that the build succeeds. Currently passing.

No new issues found in the rest of the diff. Ready to merge from a correctness standpoint.

PR #238, opened 2026-07-20
chore: remove legacy tokenized-link CV export

### LINKED ISSUE #211: [Feat] Turn on CV ZIP export + retention policy
### Description
Companies need to review student CVs for months after the event, but the current tokenized-link export depends on the site staying up and on links that can expire mid-review. CVs are PII; GDPR applies.

### Why
The tokenized-link approach has two real failure modes: it breaks if the site goes down during the review window, and bearer tokens floating in exported files are a bigger leak surface than a self-contained export. A ZIP-at-export-time download is already built and doesn't depend on the site staying up.

### How
Turn on the existing ZIP-of-CVs-at-export-time feature (Option A). Add a retention policy (3–6 month window) with auto-purge after expiry, and update the privacy policy to reflect the new retention/export model.

### Dependencies
The exact retention window (3 vs. 6 months) and the privacy-policy wording need sign-off before this ships — flagged as a team-discussion item, not a unilateral technical call.

david-fs-valente:
`DownloadButton` (CSV export with per-CV non-expiring JWT links) was already unreachable from the UI, but `/api/export`, `/api/export/[code]/cv`, and the token mechanism behind them were still live. The ZIP export (`/api/companies/history/cv-zip`) already replaces this with a self-contained download. Removes the dead component, both routes, now-unused service/repository functions, and stale AGENTS.md references.

Refs #211.


david-fs-valente:
## Non-blocking suggestions

- `DownloadIcon` in `src/styles/Icons.tsx` is now an orphaned export — after this PR removes `DownloadButton`, nothing in `src/` imports it anymore. Fine to leave for now, but worth a follow-up cleanup or noting if it's intentionally kept for future use.
- Branch is named `chore/remove-legacy-cv-export`. Per AGENTS.md, `chore/` is described as for "non-code tasks (docs, config, deps, tooling)", while this PR removes actual application code (routes, service functions, a component). `refactor/` might fit the Conventional Branch spec slightly better for a dead-code removal like this — very minor, not blocking.

Everything else checks out: verified against the actual PR head that no references to the removed symbols/routes remain anywhere in `src/`, confirmed `DownloadButton` was already unreferenced dead code before this PR (checked the parent commit), the replacement ZIP export path is untouched, `downloadjs` is correctly kept in package.json (still used by `ExcelButton`), and CI (test/typecheck/lint) is green.

PR #240, opened 2026-07-20
refactor: project structure rework

david-fs-valente:
## Summary

Implements the approved #134 proposal, in logically-scoped commits:

1. **Merge `application/domain/*` into top-level `domain/`** — collapses the duplicated "pure business rules" concept. `authPolicy`, `actionRules`, `saveRules`, `studentAccess` move into new camelCase concern-scoped folders (`domain/auth/`, `domain/action/`, `domain/savedStudent/`, `domain/student/`); the pre-existing `domain/Company/` and `domain/Student/` are renamed to `domain/company/` and `domain/student/` for casing consistency.
2. **Drain `services/`** — `authService.ts` → `application/services/` (gains the `"server-only"` marker matching its siblings); `api.ts` → `config/`; `apiResponse.ts` deleted outright (zero callers, already superseded by `lib/http/server.ts`'s `httpErrorResponse`).
3. **Fold `styles/` and `presentation/`** — `Icons.tsx` → `components/ui/` (first entry there); `tier-styling.ts` → `domain/company/services/`, alongside `company-tier-access.ts`.
4. **Docs** — adds a "where new code goes" table to AGENTS.md's Conventions section, fixes stale paths from the prior three commits, reserves a root-level `tests/e2e/` folder for future integration/smoke tests (existing colocated unit tests untouched).
5. **Adopt `components/ui/` for shared primitives** — `FileInput`, `HeadingText`, `Input`, `InputSelect`, `Modal`, `PrimaryButton`, `PrimaryLinkButton`, `PrivacyPolicyModal`, `Spinner` (9 components with 2+ consumers that are genuinely generic UI primitives, not domain/feature composites) move into `components/ui/`.
6. **Docs correction** — drops the route-local `_components/` convention from AGENTS.md; considered during this work but rejected in favor of keeping all components under `components/`.

**Explicitly out of scope for this pass** (left flat in `components/`, deliberately):
- Multi-consumer feature/layout composites (`Companies`, `Faq`, `Footer`, `HistorySection`, `Profile`, `Schedule`, `SessionAuthLayout`, `Sponsor`, `TopBar`) — reused across routes but not primitives in the design-system sense.
- 14 single-consumer components — no route-local colocation, per the above.
- 9 components confirmed to have zero references anywhere in `src/` (`Activity`, `AnimatedText`, `ComingSoon`, `DownloadButton`, `Highlight`, `HighlightInfoBit`, `Map`, `TextArea`, `TopBarLink`) — left untouched by product decision, not part of this structural refactor.

PR #241, opened 2026-07-20
feat: add admin backoffice

### LINKED ISSUE #202: [Feat] Add a full admin CRUD section (users, data, storage)
### Description
There's currently no unified admin interface for managing users, data, and storage/files — DB changes happen ad hoc (direct DB access, one-off scripts, or manual Supabase Storage cleanup) rather than through the app.

### Why
As the dataset grows edition over edition, ad hoc fixes get riskier (no audit trail, no confirmation step, whoever's available reaching for direct prod DB access) and don't scale to a team that rotates yearly.

### How
Scope a full admin CRUD section covering: user management (students/companies/employees), general data (actions, interests, saved-students), and storage/files (avatars, CVs) where feasible — reusing the existing `admin` auth Strategy (`defineHandler({ auth: "admin" })`) and the repository/service layers rather than raw Prisma calls in routes. Start by scoping which entities actually still need admin CRUD vs. what's already covered (Company/Sponsor already has an admin UI, see #152/#153).

### Dependencies
None directly, but should reuse the repository/service/HTTP-layer conventions from #126/#149 rather than introducing a new pattern.

david-fs-valente:
## Summary

Implements #202 ("Add a full admin CRUD section") in full — layout shell, every entity's CRUD, and the Company tier board — as agreed over the course of scoping this out.

**Layout shell**
- Sidebar nav on the shared `(admin)/layout.tsx` grouped by domain: Users (Students, Employees), Domain (Actions, Interests), Edition (Sponsors, Companies), Files (Storage), Statistics, Logs (external link to GlitchTip/Sentry via a new `NEXT_PUBLIC_LOGS_DASHBOARD_URL` env var).
- Centralized the `session.isAdmin` check into the layout (via `notFound()`) — previously duplicated per-page.
- Statistics page with real content from the start: total registered students, saved-students-per-company.

**Reusable building blocks**
- `DataTable`: paginated, sortable (`?sort=&order=`), searchable (`?q=`) — all URL-driven and server-renderable, no client state for any of it.
- `AdminForm`: sections API (`fields` incl. a `multiSelect` checklist kind, `image` with live preview/upload, `password` — only submitted when filled and matching) instead of one flat field list.
- `AdminToggleButton` / `AdminDeleteButton`: generic, URL/field-driven row actions (not callback props, since Server Component pages can't pass functions to Client Components).
- A new admin-gated upload route, `POST /api/admin/storage/avatar`, separate from the existing unauthenticated `storage/avatar` route (tracked by #215, not touched here).

**Per-entity CRUD**
- **Companies / Sponsors**: retrofitted off their old inline-edit-row components onto `DataTable`/`AdminForm`; active toggle, no hard delete.
- **Actions**: was list + `isLive` toggle only — added create/full-update; the QR detail page (`/actions/[id]`) is untouched.
- **Interests**: was pure read-only — added list (with computed usage count)/create/update/delete; delete blocked server-side (409) while still in use.
- **Students / Employees**: added a `User.active` migration (deactivating the User deactivates both, since they're 1:1 on the same id), `createSupabaseAuthUserAsAdmin`/`rollbackAuthUser` (admin sets the password directly at creation for both — no email-invite flow), full list/create/edit/toggle/delete, company picker on the Employee form.
- **Storage**: genuinely new — list/delete for the `avatars` (public, thumbnails) and `cvs` (private, signed URLs) buckets, at `/storage` and `/storage/cvs`.
- **Company tier board** (`/companies/tier-board`): tiermaker-style drag-and-drop (`@dnd-kit`) — one lane per tier, drag to reorder/change tier, explicit "Guardar" fires a single bulk-update transaction; never autosaves on drag.

**Also fixed along the way**: `client-only` was imported by existing code (`lib/http/client.ts`, `client/api/upload.ts`) but never actually installed as a dependency — Next's build tolerated it, Vitest didn't.


david-fs-valente:
## Blocking issues

**Deactivating a Student/Employee (`User.active`) does not actually restrict their access — it's a no-op beyond the DB flag.**

This PR adds `User.active` (migration `20260720211709_add_user_active`) with the schema comment "Gates a Student/Employee's ability to log in", and wires an `AdminToggleButton` in the Students/Employees admin pages that calls `updateStudentForAdmin`/`updateEmployeeForAdmin` → `updateUserActive(id, active)`, which only does `prisma.user.update({ data: { active } })`.

Nothing in this diff (or in the unmodified `src/application/services/sessionService.ts`, which `getServerSession()` lives in, or `src/application/domain/authPolicy.ts`) ever reads `user.active`. There's also no corresponding Supabase Auth-side action (e.g. `admin.auth.admin.updateUserById(id, { ban_duration: ... })`) to actually lock the account out at the auth layer.

Net effect: an admin toggles a student/employee to "inactive" in the backoffice, gets a UI that looks like it revoked access, but the user can still log in via Supabase Auth and `getServerSession()` will resolve them exactly as before (no `active` check anywhere in the session/auth chain) — full read/write access to everything a normal active account has. This is a functional bug that's actively misleading for an access-control feature — please either:
- add the `active` check to `getServerSession()` (and/or `passesAuthPolicy`) so a deactivated user's session actually gets rejected, or
- ban the Supabase Auth user directly when `active` is set to `false`,

and add a test covering that a deactivated user's session/route access is actually denied.

david-fs-valente:
## Non-blocking suggestions

- **`src/application/services/storageAdminService.ts` (`deleteStorageObject`) / `src/app/api/admin/storage/[type]/[name]/route.ts`**: `name` is taken from the `[name]` dynamic segment, `decodeURIComponent`'d, then interpolated directly into `${prefix}/${name}` for `.remove([...])`. A `name` containing an encoded `../` could delete objects outside the intended `distribution/avatar`/`distribution/cv` prefix within the same bucket. Low severity since the route is `auth: "admin"`-gated (trusted caller), but worth normalizing/rejecting path separators defensively, especially since the `cvs` bucket holds private data.

- **Test coverage**: the only new tests are `AdminForm/index.test.tsx` and `DataTable/index.test.tsx` (both UI). None of the new service-layer business rules got colocated tests, e.g.:
  - `employeeService.createEmployeeForAdmin` / `studentService.createStudentForAdmin`'s rollback-on-failure path (Supabase auth user created, then Prisma transaction fails → `rollbackAuthUser` called)
  - `interestService.deleteInterestForAdmin`'s 409-when-in-use rule
  - `companyService.updateCompanyTierBoard`'s bulk transaction
  
  Per AGENTS.md's "extend the relevant test file when behavior changes" convention, these would be good candidates given they encode real business rules, not just plumbing.

- **`rollbackAuthUser`**: if the Supabase `deleteUser` rollback call itself fails (network blip, etc.), the error is only reported via `reportError` — the admin-facing create call still surfaces the original error, with no indication that a now-orphaned Supabase Auth user (with no matching Prisma `User`) was left behind. Not a big deal operationally, but worth a code comment noting the known gap (similar to how other known gaps are documented elsewhere in this codebase) so it's not mistaken for full transactional safety.

- **`src/app/api/admin/storage/avatar/route.ts`**: comment above it explains why it isn't a `defineHandler` route (multipart body) and why it's separate from the existing unauthenticated `storage/avatar` (tracked by #215) — good, just flagging that the manual `if (!session || !session.isAdmin)` check here duplicates the `defineHandler({ auth: "admin" })` policy by hand; low risk since it's clearly written out, but a small refactor (e.g. a shared `requireAdmin(session)` helper) would keep the two "admin" checks from drifting if the policy ever changes.

david-fs-valente:
## Rebased + blocking issue fixed + careful review, given this manages real accounts, private CV data, and links out to GlitchTip

**Merged \`dev\` in** (3 conflicts: \`package.json\`, \`pnpm-lock.yaml\` regenerated from the merged manifest, \`studentService.ts\`'s domain imports). Also caught two silent (non-conflicting) breaks git didn't flag: 3 stale \`@/domain/Student/\`\`@/domain/Company/\` casing references left over from #240's refactor, and a duplicate \`deleteUser()\` declaration in \`userRepository.ts\` - this branch and dev's AuthNEI hardening fix each added their own independently, in non-overlapping parts of the file.

## Blocking issue fixed (\`8d217f6\`) - \`User.active\` now actually gates access

\`getServerSession()\` now rejects the session outright for an inactive \`User\`, on both the id and email lookup paths - a single enforcement point every auth-gated route/page already goes through. Also bans the underlying Supabase Auth user (\`ban_duration\`, best-effort) when an admin toggles active off, and unbans on toggling back on, as defense-in-depth on top of the DB flag. 11 new tests across \`sessionService\`, \`studentService\`, \`employeeService\`, and \`authApplicationService\`.

Side finding while wiring this up: \`sessionService.ts\` had **zero test coverage** because \`react@18.3.1\` (the version actually installed) doesn't export \`cache\` - Next's bundler substitutes a build that does for server-only code, Vitest doesn't, so any test importing it threw \`"cache is not a function"\`. Stubbed as a pass-through in \`vitest.setup.ts\` (safe - \`cache()\` is a per-request memoization optimization, not a correctness dependency), which should also unblock testing this file's other callers going forward.

## Careful review - what I checked

- **Every admin route's auth gating** - all 16 routes are \`defineHandler({ auth: "admin" })\` except the one multipart upload route, which has an equivalent manual \`session.isAdmin\` check. Consistent, no gaps found.
- **Sort/search query construction** - every admin-listed entity (student, employee, company, sponsor, action, interest) validates \`sort\` against an explicit field allowlist before it reaches Prisma's \`orderBy\`, and \`search\` goes through Prisma's parameterized \`contains\` filter. No injection surface.
- **Company tier board bulk update** - wrapped in \`prisma.$transaction()\`, atomic; \`z.uuid()\`/enum-validated input.
- **Statistics page** - simple, read-only, protected by the same layout-level admin gate as everything else.
- **GlitchTip logs link** (\`NEXT_PUBLIC_LOGS_DASHBOARD_URL\`) - Zod-validated as a URL, conditionally rendered, correct \`target="_blank" rel="noopener noreferrer"\`. No issues.
- **Fixed (\`e5308a6\`):** the previously-flagged path-traversal note in admin storage delete - \`deleteStorageObject()\` interpolated the decoded \`[name]\` segment directly into the object key. Low severity (admin-gated, and the UI only ever passes back names it just listed), but the \`cvs\` bucket holds private data, so added an allowlist regex (legitimate names are always a bare \`uuidv4()\`, optionally \`.pdf\`-suffixed - never a path separator) plus 4 tests.
- **Admin-set password validation** (\`min(8)\`, no complexity requirement) - checked against the existing self-signup schemas (\`signUpSchema.ts\`, \`employeeSignUpSchema.ts\`): this is consistent with the existing app-wide policy, not a regression introduced here. The complexity requirement you'll see in the signup UI is client-side only, pre-existing, out of scope for this PR.
- **Interest deletion 409-in-use rule**, **toggle/delete button wiring**, **createSupabaseAuthUserAsAdmin rollback path** - read through, logic is correct as documented in the earlier review comment.

CI green, \`mergeStateStatus: CLEAN\`. The remaining non-blocking suggestions from the earlier review (more service-layer test coverage, the \`rollbackAuthUser\` orphan-on-failure comment, the duplicated admin-check-vs-\`defineHandler\` nit) are still open but genuinely non-blocking - your call whether to fold those into this PR or track separately.

PR #243, opened 2026-07-21
build: add BuildKit cache mount for pnpm store in Dockerfile

### LINKED ISSUE #205: [Chore] Add BuildKit cache mount for pnpm store
### Description
Docker builds take ~5 minutes, mostly driven by installing `node_modules` (~1.4GB — native binaries like `sharp`, Prisma engines, Sentry tooling). The Dockerfile's `deps` stage runs a plain `pnpm install --frozen-lockfile` with no persistent cache, so every deploy re-downloads all dependencies from scratch even when only one package changed.

### Why
We're a non-profit running everything on a single self-hosted VPS — every extra minute of build time is CPU/bandwidth spent that doesn't need to be, and re-downloading unchanged deps on every deploy is pure waste. A BuildKit cache mount needs no new infra or service, it's a free win.

### How
Add a BuildKit cache mount (`--mount=type=cache,id=pnpm,target=<pnpm store path>`) to the `deps` stage's install step in `Dockerfile`. pnpm's store is content-addressable, so this persists across builds on Coolify's host and only fetches new/changed packages going forward — existing deps are reused as-is. Requires builds to run on the same persistent Docker host, already true for a normal self-hosted Coolify setup.

### Dependencies
None directly, but touches the same Dockerfile stages as #206 (migrate build split) — worth sequencing together or in the same PR.

david-fs-valente:
## Summary
- Docker builds currently run a plain `pnpm install --frozen-lockfile` in the `deps` stage with no persistent cache, so every deploy re-downloads all of node_modules (~1.4GB, including native binaries like sharp, Prisma engines, and Sentry tooling) from scratch even when only one package changed.
- Adds a BuildKit cache mount (`--mount=type=cache,id=pnpm,target=/pnpm/store`) to the `deps` stage's install step, combined with `pnpm config set store-dir /pnpm/store` so pnpm's content-addressable store actually persists at that mounted path across builds on the self-hosted VPS (Coolify), instead of re-fetching unchanged packages every deploy.
- No other stages were restructured — this only touches the `deps` stage's install `RUN` step.

Closes #205

## Verification
- `pnpm lint` — clean, no issues in touched files.
- `pnpm typecheck` — clean.
- `docker build --target deps` — builds successfully.
- Manually verified the cache mount actually works: forced the install layer to re-run (via a temporary cache-busting `ARG`, since layer invalidation is otherwise needed to observe reuse) while keeping the persistent `/pnpm/store` mount intact. Result: pnpm reported `reused 1146/1148` packages, `downloaded 0`, and the install step finished in ~15s versus ~75-90s on a cold store. Confirms the intended bandwidth/build-time savings on repeat builds where the lockfile is largely unchanged.
- Did not measure real-world build-time improvement on the actual Coolify/VPS host — that depends on the BuildKit cache directory persisting there across deploys, which this PR can't verify locally.

## Note for reviewers
Per issue #205's "Dependencies" note, this shares the `deps`/`builder` Dockerfile stages with issue #206 (migrate build split), so it may be worth reviewing together. #206 itself is out of scope for this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,10 @@ COPY prisma ./prisma
 # BuildKit cache mount lets pnpm skip re-downloading unchanged packages —
 # `store-dir` is pinned explicitly so pnpm actually writes into the mounted
 # path instead of whatever its platform default would otherwise resolve to.
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
+# `sharing=locked` serializes access to the mount so two overlapping builds
+# on the same host (e.g. Coolify deploying two branches at once) can't race
+# on writes to the store.
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store,sharing=locked \
   if [ -f pnpm-lock.yaml ]; then \
     pnpm config set store-dir /pnpm/store && \
     pnpm install --frozen-lockfile; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,14 @@ WORKDIR /app
 COPY package.json package-lock.json* pnpm-lock.yaml* ./
 COPY prisma ./prisma
 
-# Install dependencies based on the preferred package manager
-RUN \
+# Install dependencies based on the preferred package manager.
+# The pnpm store is content-addressable, so persisting it across builds via a
+# BuildKit cache mount lets pnpm skip re-downloading unchanged packages —
+# `store-dir` is pinned explicitly so pnpm actually writes into the mounted
+# path instead of whatever its platform default would otherwise resolve to.
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
   if [ -f pnpm-lock.yaml ]; then \
+    pnpm config set store-dir /pnpm/store && \
     pnpm install --frozen-lockfile; \
   elif [ -f package-lock.json ]; then \
     npm ci; \


david-fs-valente:
## Non-blocking suggestions

- **Cache mount concurrency:** the mount uses the default `sharing=shared` mode. That's fine for pnpm's content-addressable store in the common case (single serialized deploy at a time), but if Coolify ever triggers overlapping builds on the same host (e.g. two branches deploying concurrently), consider `sharing=locked` to rule out any store-write races. Not urgent given this team's deploy cadence — just flagging for awareness.
- **BuildKit-on-Coolify assumption:** the PR description notes real-world persistence on the VPS couldn't be verified locally, which is reasonable to merge without. Worth a quick post-merge check that a couple of consecutive deploys actually show the `reused X/Y packages` speedup in the Coolify build logs — since the `deps` layer is skipped entirely via normal Docker layer caching whenever the lockfile is unchanged, the cache-mount benefit is only observable on lockfile-changing deploys. (Side note, not a new risk: the Dockerfile already depends on BuildKit today via the existing `--mount=type=secret` for `SENTRY_AUTH_TOKEN` in the `builder` stage, so this PR isn't introducing a new hard dependency on BuildKit — just confirming the assumption holds is worthwhile.)

PR #244, opened 2026-07-21
feat: add AuthNEI (Zitadel) OAuth sign-in option for students

### LINKED ISSUE #235: [Feat] Add AuthNEI (Zitadel) OAuth sign-in option for students
### Description
Add "Continuar com AuthNEI" as an additional sign-in/sign-up option for students, alongside the existing Supabase email/password flow — not a replacement. AuthNEI (`../authnei`) is NEI's self-hosted Zitadel OIDC identity provider, already used as the sole login mechanism by `../antirecurso` (via NextAuth). This app uses Supabase Auth directly (not NextAuth), so the integration point here is Supabase's native custom OIDC provider support rather than antirecurso's NextAuth pattern.

### Why
Students already carry (or will carry, per the org's shared-IdP push) an AuthNEI identity for other NEI apps — letting them use it here removes the need to invent another password and re-prove an institutional identity ISEP already vouches for at the AuthNEI layer, while keeping local email/password unchanged for employees, admins, and anyone without an AuthNEI account. `../authnei`'s onboarding checklist and `sub`-based canonical identity model already give us a documented path for registering this app as a client.

### How
- Register a new Zitadel application in AuthNEI (Web app, Authorization Code + PKCE, scopes `openid email profile`) following `../authnei`'s onboarding checklist — dev and prod get separate app registrations.
- Configure a Supabase **custom OIDC provider** (`custom:authnei`) via the dashboard: OIDC auto-discovery against AuthNEI's issuer URL, client ID/secret from the Zitadel app above. This supersedes the older "register it as Keycloak" workaround some teams use — Supabase now supports generic custom OIDC providers directly.
- Add a new `src/app/auth/callback/route.ts` to exchange the OAuth `code` for a session (`exchangeCodeForSession`). The existing `src/app/auth/confirm/route.ts` only handles OTP-based email-link confirmation, not the OAuth code flow, so it can't be reused as-is.
- UI: add a "Continuar com AuthNEI" button above the existing form on `/login`, and as a fast-path option on step 1 of the student signup wizard (`StudentSignUp`) that skips manual ISEP-email + password entry in `AccountDetailsStep` (year selection still applies). The Bio and Interests steps still run afterward, since AuthNEI only proves identity, not our domain-specific profile data. Employee signup (`/signup/employee`) is unaffected.
- **Before enabling this, whoever picks it up must check the hosted Supabase project's dashboard** (Authentication → Providers → Email) and confirm whether **"Confirm email" is turned on**. Supabase auto-links a new OAuth identity into an existing account by matching email, and it only strips out a dangling *unconfirmed* local password identity before doing so — if "Confirm email" is off, an existing (or attacker-created) unconfirmed account sitting on a spoofed ISEP email would silently absorb the real student's AuthNEI login, handing standing account access to whoever created that local account first. `supabase/config.toml`'s `enable_confirmations = false` only applies to the local CLI dev stack, not the hosted project, so this can't be verified from the repo — it must be checked directly in the dashboard, and turned on before shipping if it isn't already.

### Dependencies
None directly. Depends on `../authnei` having a registered project/app for this repo's dev and prod environments.

david-fs-valente:
## Summary

Adds "Continuar com AuthNEI" as an additional student sign-in/sign-up option, alongside (not replacing) the existing Supabase email/password flow. AuthNEI is NEI's self-hosted Zitadel OIDC identity provider, integrated via Supabase's native custom OIDC provider support (`custom:authnei`).

- `src/app/auth/callback/route.ts` — new OAuth callback route: exchanges the `code` query param for a session via `exchangeCodeForSession`, then routes the user based on whether they already have an app profile (existing student → their profile, existing employee → `/dashboard`, brand-new identity → provisioned as a `STUDENT` `User` row and sent into the signup wizard). Lives outside `(auth)`/`api`, mirroring `app/auth/confirm/route.ts`'s precedent (both are Supabase-constructed callback URLs).
- `src/application/services/authApplicationService.ts` — new `completeOAuthSignIn()` service function backing the callback route (upserts the Prisma `User` row for first-time AuthNEI identities; no password involved since Supabase manages the OAuth identity itself).
- `src/components/AuthNeiButton` — new reusable client component wrapping `supabase.auth.signInWithOAuth({ provider: "custom:authnei", ... })`, with a toast on failure.
- `src/app/(auth)/login/page.tsx` — AuthNEI button added above the existing email/password form.
- `src/components/SignUp/AccountDetailsStep` + `src/components/StudentSignUp` — AuthNEI fast-path on step 1: skips manual ISEP-email + password entry, restores in-progress wizard data (name) via `sessionStorage` across the OAuth redirect, and resumes at year selection. Bio and Interests steps still run afterward, since AuthNEI only proves identity, not domain profile data.
- `src/client/api/auth.ts` — `signUp()` now skips the `/auth/signup` (account-creation) call when no password is present, since AuthNEI signups already have a Supabase auth identity and Prisma `User` row from the callback.
- `src/config/index.ts` — `custom:authnei` provider id added as a plain string constant (not a secret, so no new env var).

Employee signup (`/signup/employee`) is untouched.

## Manual steps required before this works

- [x] Register the Zitadel app in AuthNEI (Web app, Authorization Code + PKCE, scopes `openid email profile`), separately for dev and prod, per `../authnei`'s onboarding checklist.
- [x] Configure Supabase's custom OIDC provider `custom:authnei` via the dashboard (OIDC auto-discovery against AuthNEI's issuer URL, client ID/secret from the Zitadel app).
- [x] **Security-critical:** before enabling, confirm the hosted Supabase project's dashboard (Authentication → Providers → Email) has "Confirm email" **ON**. If it's off, an unconfirmed local account sitting on a spoofed ISEP email could silently absorb a real student's AuthNEI login on first sign-in (Supabase auto-links OAuth identities to existing accounts by matching email, and only strips a dangling *unconfirmed* local password identity before doing so). `supabase/config.toml`'s `enable_confirmations = false` only applies to the local CLI dev stack and does **not** reflect the hosted project's actual setting — this cannot be verified from the repo and must be checked directly in the dashboard.

## Test plan

- [x] `pnpm lint` — clean
- [x] `pnpm typecheck` — clean
- [x] `pnpm test` — 60/60 passing (22 files)
- [x] `pnpm dev` against a local ephemeral Postgres (no real Supabase available in this environment): `/login` renders with the new button alongside the untouched email/password form; `/signup` renders normally; `/auth/callback` (no `code`, and a fake `code` against an unreachable Supabase URL) both redirect cleanly to `/auth/auth-code-error` (307) instead of crashing — confirmed in server logs the `exchangeCodeForSession` network failure (`ECONNREFUSED`) is caught, not thrown.
- [x] **Not verified — cannot be, in this environment:** the actual end-to-end OAuth exchange (real AuthNEI login → Supabase code exchange → session → wizard resume) requires the real AuthNEI + Supabase dashboard configuration listed above. Clicking the button in a live browser against a real Supabase project, and the sessionStorage draft-restore/step-resume behavior in `StudentSignUp`, were verified by code review only, not a live click-through (no browser automation available in this sandboxed environment).

Refs #235


david-fs-valente:
## Blocking issues

**1. Open redirect in `sanitizeNext()` via backslash normalization** — `src/app/auth/callback/route.ts`

```js
function sanitizeNext(next: string | null): string {
  if (!next || !next.startsWith("/") || next.startsWith("//")) return "/";
  return next;
}
```

This blocks `//evil.com` but not a backslash-prefixed value. WHATWG URL parsing normalizes backslashes to forward slashes for special schemes (http/https), so:

```js
new URL("/\\evil.com", "https://example.com").href
// => "https://evil.com/"
```

`next=/\evil.com` (or its `%5C`-encoded form) passes `sanitizeNext` unchanged (it starts with `/`, not `//`), and `NextResponse.redirect(new URL(destination, request.url))` then resolves it to an external host. This is reachable without needing the app's own `AuthNeiButton` UI: `next` comes straight from the incoming callback URL's query string, and the redirect that lands on `/auth/callback` is fully attacker-constructible via a direct call to `supabase.auth.signInWithOAuth({ redirectTo })` (public anon key, client-side). A victim completes a genuine AuthNEI login (nothing looks wrong in the browser during that step) and is silently bounced to an attacker-controlled page at the end — classic post-login open-redirect/phishing setup, and exactly what the code's own comment says it's trying to prevent ("only allow redirecting back into this app, never to an external host").

Fix: reject any `next` containing a backslash (or normalize/strip `\` before the prefix checks), not just `//`.

**2. Existing password/email students hit a broken flow when using AuthNEI with their existing institutional email** — `completeOAuthSignIn()` in `src/application/services/authApplicationService.ts`

```ts
const existing = await findUserSessionById(input.id);
if (existing) { ... }
await upsertUser({ id: input.id, email: input.email, role: "STUDENT" });
```

Lookup is keyed only on the Supabase auth `id`, never falling back to matching by email. `email` is `@unique` on the Prisma `User` model. Per the PR's own security note, Supabase only auto-links an OAuth identity to an existing account when that existing account is a *dangling unconfirmed* one — a normal, already-confirmed student who signed up earlier with email/password will **not** be auto-linked, so AuthNEI issues a distinct Supabase auth `id` for the same institutional email. `findUserSessionById(input.id)` then finds nothing, falls through to `upsertUser`'s `create` branch, and violates the unique constraint on `email`. The thrown error is swallowed by the route's `try/catch` into a generic `reportError` + redirect to `/auth/auth-code-error`, silently locking that student out of the new button with their own email instead of signing them into their existing account.

This isn't a rare edge case — it's the primary path for any student who registered before this feature shipped and then tries the new "Continuar com AuthNEI" option. Needs a fallback lookup by email (and a real linking/merge decision) before this ships.

david-fs-valente:
## Non-blocking suggestions

- **No colocated tests for the new auth-flow logic.** `completeOAuthSignIn()` (branching: existing student / existing employee / STUDENT-role-but-no-profile / brand-new identity) and `sanitizeNext()` in the new callback route are both pure, easily-unit-testable functions with no test coverage added, despite the repo's Vitest auto-discovery convention ("extend the relevant test file when behavior changes"). A `sanitizeNext` test would likely have caught the backslash open-redirect from the blocking comment.
- **`data.user.email as Email` bypasses the `Email` branded type's validation** (`src/app/auth/callback/route.ts`). Everywhere else in the codebase an `Email`-typed value is expected to have passed through `Email.create()`/`Email.createInDomain()` (length limits, format regex, dot rules). This is a raw type assertion instead, so a value from a misconfigured/unusual OIDC attribute mapping could carry an `Email` type without actually satisfying those invariants downstream. Consider `Email.create(data.user.email)` (with the throw caught into the existing error-redirect path) instead of the cast.
- **AuthNEI path skips the `isIsepEmail` domain check** that `src/schemas/signUpSchema.ts` enforces for the manual password signup path (`EmailSchema.refine(isIsepEmail, ...)`). Presumably intentional — AuthNEI enrollment itself is assumed to be ISEP-gated — but worth confirming explicitly rather than leaving it as an implicit divergence between the two signup paths for the same `STUDENT` role.
- **`completeOAuthSignIn`'s `EMPLOYEE` branch returns `/dashboard` unconditionally**, without checking `existing.employee` truthiness the way the `STUDENT` branch checks `existing.student`. Harmless today since AuthNEI is only wired up for students, but inconsistent if this provider is ever reused for employee sign-in.
- The PR body itself lists an unchecked, security-critical manual step ("confirm 'Confirm email' is ON in the hosted Supabase dashboard") — not something fixable in this diff, but flagging so it isn't forgotten before this is enabled against the real Supabase project.
- Minor DRY nit: the "ou" divider markup (`h-px flex-1 bg-[rgba(255,255,255,0.2)]` pair) is duplicated between `login/page.tsx` and `AccountDetailsStep`; could be factored into a small shared component if a third use case shows up.

david-fs-valente:
## Re-review update

Both blocking issues from the previous review are fixed, and the branch is now rebased cleanly on `dev` (CI green, `mergeStateStatus: CLEAN`).

**1. Open redirect via backslash in `sanitizeNext()` — fixed (`42a4640`).** Backslashes are now rejected outright before the `/`/`//` checks, closing the `/\evil.com` bypass (WHATWG URL parsing normalizes a leading `/\` to `//` for http(s) URLs). Moved into its own module (`sanitizeNext.ts`) since Next.js's App Router typegen rejects extra named exports from `route.ts`, with 5 new unit tests covering the backslash, `//`, missing, and valid-path cases.

**2. Existing password/email students silently locked out — fixed (`0589649`).** `completeOAuthSignIn()` now falls back to an email lookup when the Supabase auth id isn't found, and re-keys the existing `User` row onto the new AuthNEI id instead of hitting the `email` unique constraint. This is safe because `Student.id`, `Employee.id`, and the `Interest` join table's `User` FK are all `ON UPDATE CASCADE` at the DB level, so the student's profile, action completions, and saved-by-employee rows carry over intact. Also swapped the raw `data.user.email as Email` cast for `Email.create()`, and added the `existing.employee` truthiness check to the `EMPLOYEE` branch for consistency. 6 new unit tests cover all `completeOAuthSignIn` branches.

## New finding from re-review: the relink fix's safety fully depends on "Confirm email" being ON

Tracing through the attack surface once more: `completeOAuthSignIn`'s new email-fallback match is intentionally permissive — it matches *any* existing `User` row by email, regardless of whether that account was ever actually confirmed. That's necessary for the fix's own purpose (a legitimately-confirmed pre-AuthNEI student must be found and relinked), but it also means an *unconfirmed* row matches just as readily.

`signUpUser()` → `upsertUser()` already creates the Prisma `User` row immediately on `/auth/signup` (a public, unauthenticated endpoint gated only by `isIsepEmail`'s format check, not real ownership), regardless of email-confirmation status. So an attacker can already plant a dangling `User` row under any syntactically valid ISEP email today, confirmed or not — that part isn't new.

What determines whether this is harmless or exploitable is whether the attacker's spoofed signup ever gets an active **session**: `/students` (where a `Student` profile with attacker-controlled bio/avatar/CV/interests would actually get created) requires `auth: "session"`. Supabase only grants a session immediately post-`signUp()` when **"Confirm email" is OFF**. With it ON, the attacker's row stays dangling and profile-less, and the relink is harmless — the real student just lands in a normal, empty signup wizard under their own (relinked) id. With it OFF, the attacker gets an authenticated session immediately, can build out a fully attacker-controlled profile under the spoofed email, and my relink logic would then hand that poisoned profile to the real student the moment they use AuthNEI.

So this doesn't need a new code fix by itself — but it means the PR's existing unchecked manual step ("confirm 'Confirm email' is ON in the hosted Supabase dashboard before enabling") is now a **hard precondition for this relink logic to be safe**, not just a mitigation for Supabase's own auto-link edge case as originally scoped. I'd recommend also adding a server-side confirmation check (verify the existing identity's `email_confirmed_at` via the Admin API before trusting the relink) as defense-in-depth, so this doesn't rest solely on a dashboard toggle staying correctly configured forever — happy to add that in a follow-up if wanted.

david-fs-valente:
## Defense-in-depth added (\`da125ad\`)

Per the finding above, added a confirmation check before trusting the email-fallback relink: \`completeOAuthSignIn()\` now verifies the existing identity's \`email_confirmed_at\` via the Supabase Admin API (\`getUserById\`) before relinking onto it.

- **Confirmed** → relinked as before.
- **Unconfirmed** → treated as a dangling/spoofed placeholder and discarded (\`deleteUser\`, cascading to any \`Student\`/\`Employee\`/interests) instead of relinked onto, so AuthNEI claims the email fresh.
- **Admin lookup itself errors** → fails closed (treated as unconfirmed), logged via \`reportError\`.

5 new/updated unit tests cover the confirmed-relink, unconfirmed-discard, and fail-closed-on-error paths. This is defense-in-depth on top of the "Confirm email" dashboard setting, not a replacement for it — that setting still gates whether a spoofed pre-registration can ever get an active session and build out a profile in the first place, so it's still a hard precondition to verify before enabling AuthNEI in production (tracked separately).

CI green, still cleanly mergeable into \`dev\`.

PR #245, opened 2026-07-22
feat: add in-memory rate limiting to upload endpoints

### LINKED ISSUE #213: [Feat] Add in-memory rate limiting to upload endpoints
### Description
No rate limiting exists on any endpoint today. Login/signup is already covered by Supabase's own limits, but the app's own file upload endpoints (CV, avatar) aren't protected at all.

### Why
Upload endpoints are the real gap — unlimited file uploads are a resource-exhaustion/abuse vector with no protection today, and it's the one class of endpoint Supabase's built-in limits don't reach.

### How
Add simple in-memory rate limiting to the CV and avatar upload endpoints. No new infrastructure needed since the app runs on a single server. Revisit only if the app ever scales to multiple servers, since in-memory state wouldn't be shared across instances.

### Dependencies
None.

david-fs-valente:
## Summary
- Add an in-memory fixed-window rate limiter (`src/lib/rateLimit.ts`) keyed by client IP
- Apply it to the CV and avatar upload endpoints (`/api/storage/cv`, `/api/storage/avatar`), the one abuse vector Supabase's own auth rate limits don't cover
- 5 uploads/minute per IP per endpoint; exceeding it returns `429` with a `Retry-After` header
- No new infrastructure (Redis, etc.) — the app runs on a single server today, so a shared store isn't needed; revisit only if that changes

Closes #213


david-fs-valente:
## Blocking issues

1. **`getClientIp()` trusts the client-supplied `X-Forwarded-For` value, making the rate limiter trivially bypassable.** In `src/lib/rateLimit.ts`, the key is derived from `forwardedFor.split(",")[0]` — the left-most entry. Standard reverse proxies (nginx, Traefik/Coolify) typically *append* the real client IP to any pre-existing `X-Forwarded-For` header rather than overwrite it, so an attacker can send a fresh, arbitrary `X-Forwarded-For: <random>` header on every request and get a brand-new bucket each time — completely defeating the limiter. That's the exact abuse vector this PR (closes #213) is meant to close, so as written it doesn't actually stop a deliberate attacker, only accidental/naive reuse.
   - This also compounds into a minor memory-growth angle: an attacker rotating spoofed IPs can grow the in-memory `buckets` Map with new entries faster than the periodic sweep (every `windowMs` = 60s) clears them.
   - Suggested fix: take the right-most (proxy-appended, trusted) hop instead of the left-most one, or confirm the deployment's reverse proxy is configured to strip/overwrite any client-supplied `X-Forwarded-For` before the app sees it — and record that assumption in the code comment, since it's otherwise a silent trust boundary.

david-fs-valente:
## Non-blocking suggestions

1. No test exercises the actual route wiring — the new `rateLimit.test.ts` covers `createRateLimiter`/`getClientIp` directly, but nothing hits `/api/storage/avatar` or `/api/storage/cv` to confirm the 429 status code and `Retry-After` header come through correctly end-to-end.
2. `rateLimit.test.ts` casts a raw `Request` via `as never` to satisfy `getClientIp`'s `NextRequest` parameter type. Narrowing the parameter to something like `Pick<NextRequest, "headers">` would let the test pass a plain `Request` without the escape-hatch cast.
3. The 429 response block (status, `Retry-After` header, `{ error }` body) is duplicated verbatim between the avatar and CV routes. If a third upload endpoint ever needs rate limiting, consider factoring the "check + respond" pattern into a small shared helper (e.g. alongside `createRateLimiter` in `lib/rateLimit.ts`) to avoid a third copy.

PR #246, opened 2026-07-22
feat: support self-hosted Sentry-compatible backends for source maps

### LINKED ISSUE #214: [Feat] Self-host GlitchTip as a Sentry SaaS replacement
### Description
We currently use Sentry SaaS (`@sentry/nextjs`). The objective is to move to a self-hosted, open-source alternative to reduce reliance on paid vendor SaaS and avoid usage-based overages, while keeping data residency control — relevant given the CV/PII handling already in place.

### Why
We're a non-profit — avoiding a recurring SaaS bill matters here the same way it does for the Docker build-cost work (#205/#206). GlitchTip accepts a drop-in Sentry-compatible DSN, so this doesn't require rewriting any `captureException`/SDK call sites, only re-pointing the DSN at a self-hosted instance. It's far lighter than Sentry's own self-hosted stack (which needs Kafka, ClickHouse, Postgres, Redis, Zookeeper) and was chosen over Bugsink for feature parity (GlitchTip covers performance/uptime monitoring, not just bare error tracking) and a standard OSS license.

### How
Deploy GlitchTip via Docker Compose alongside the existing app/DB/Coolify stack on the VPS. Point `NEXT_PUBLIC_SENTRY_DSN`/`SENTRY_ORG`/`SENTRY_PROJECT` and the source-map upload token at the self-hosted instance instead of Sentry SaaS — no SDK code changes needed since GlitchTip speaks the same protocol. Re-verify the existing PII scrubbing config (see OBSERVABILITY.md) still applies correctly against the new backend before cutting traffic over. Update OBSERVABILITY.md to describe the self-hosted setup instead of Sentry SaaS.

### Dependencies
None directly, but shares VPS resource-budget concerns with #205/#206 (Docker build cost work) — worth being mindful of combined resource usage on a single VPS running the app, DB, Coolify, and now GlitchTip.

david-fs-valente:
## Summary
- Add `SENTRY_URL` support so source-map upload during `next build` can target a self-hosted Sentry-compatible backend (e.g. GlitchTip) instead of always defaulting to sentry.io
- The DSN alone doesn't control this: it only routes runtime error events. Source-map upload is a separate build-time step with its own upload target, which previously had no way to be redirected
- Threaded through `next.config.js` (`sentryUrl` key on `withSentryConfig`, verified against the installed `@sentry/nextjs@10.65` type definitions), `Dockerfile` (ARG/ENV), `docker-compose.app.yml` (build arg), `.env.example`, and `OBSERVABILITY.md` (env table, both build examples, a new troubleshooting entry)
- No SDK code changes needed beyond this: `sentry.server.config.ts`, `sentry.edge.config.ts`, and `src/instrumentation-client.ts` already key off `NEXT_PUBLIC_SENTRY_DSN` alone, which is backend-agnostic
- Fully backward compatible: `SENTRY_URL` unset keeps defaulting to sentry.io, unchanged for anyone not using a self-hosted backend

Part of #214

## Test plan
- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean on touched files
- [x] `pnpm test` — all 60 tests pass (no runtime `src/` behavior changed, so no new tests needed)
- [x] Sanity-loaded `next.config.js` directly with `SENTRY_URL` unset to confirm no crash/behavior change for the sentry.io path
- [x] Not verified: an actual source-map upload against a live self-hosted instance — no GlitchTip deployment exists yet (see comment below for the remaining ops work)


david-fs-valente:
## Remaining work: Coolify/ops side (not code)

This PR only covers the repo's env-var plumbing. Deploying and cutting over is a separate, non-code track:

1. **DNS** — add an A/CNAME record for `glitchtip.fallstack.nei-isep.org` pointing at the VPS IP (no wildcard exists for this zone today, confirmed).
2. **Coolify** — set that hostname as the domain on the GlitchTip resource (already created from a Coolify template, not yet deployed) and deploy. Coolify/Traefik should auto-issue a Let's Encrypt cert once DNS resolves.
3. **GlitchTip account** — GlitchTip ships with no pre-seeded credentials; the first visit shows a registration page, and the first account registered becomes the instance/org owner. Sign up once the domain is reachable.
4. **Org/project setup** — inside GlitchTip, create the organization + project and collect: DSN, org slug, project slug, and an API token for source-map upload (verify its permission-scope naming in GlitchTip's own UI — not confirmed to match Sentry's `org:ci` naming 1:1).
5. **Wire env vars into Coolify** for the app's `web` service:
   - `NEXT_PUBLIC_SENTRY_DSN` → GlitchTip project DSN
   - `SENTRY_URL` → `https://glitchtip.fallstack.nei-isep.org/`
   - `SENTRY_ORG` / `SENTRY_PROJECT` → the slugs from step 4
   - `SENTRY_AUTH_TOKEN` → the token from step 4 (BuildKit secret / Coolify secret, never a plain build arg)
6. **Re-verify PII scrubbing** (`OBSERVABILITY.md` §7) against the live GlitchTip backend before enabling production alerts — the app-side scrubbing (`beforeSend`/`beforeSendLog`) is backend-agnostic, but GlitchTip's own server-side scrubbing settings need to be configured independently.
7. **Smoke-test Sentry Logs ingestion** — confirm GlitchTip actually accepts the `enableLogs`/`pinoIntegration` structured-log events this app sends, not just Issues. Not verified against GlitchTip specifically.
8. Once confirmed working end-to-end, update `OBSERVABILITY.md`'s "Create the Sentry project" section with any GlitchTip-specific menu-path notes learned in practice (currently kept generic since exact UI wasn't verified during this PR).

Resource-budget reminder from the original issue: keep an eye on combined RAM/CPU with GlitchTip running alongside the app, DB, and Coolify itself on one VPS.

Issue #203, opened 2026-07-19
[Refactor] Move timetable/schedule data into the database


david-fs-valente:
### Description
Timetable/schedule data currently lives in `src/edition/` as hardcoded config (`ScheduleDays`), not in the database — updating it requires a code change and a deploy.

### Why
Schedule details (room changes, time shifts, added sessions) often firm up close to the event; needing a deploy to fix a typo or move a slot is slower and riskier than editing a DB row through an admin UI.

### How
Add a `Schedule`/`ScheduleDay` model (Prisma migration), migrate the current `edition/` schedule content into seed data, and update the page(s) that render the schedule to read from the DB instead of the static config. Pairs naturally with #202 (admin CRUD) for actually editing it without a deploy.

### Dependencies
None required, but most useful once #202 (admin CRUD section) exists to edit the data through.

PR #247, opened 2026-07-23
feat: move FAQ data into the database

### LINKED ISSUE #204: [Refactor] Move FAQ data into the database
### Description
FAQ content currently lives in `src/edition/` as hardcoded config, not in the database — same pattern as the timetable (see #203).

### Why
FAQ answers change based on questions actually asked at the event or policy tweaks (e.g. the CV retention policy, once the CV storage decision lands) — DB-backed content means those edits don't need a deploy.

### How
Add an `FaqEntry` model (Prisma migration), migrate the current `edition/` FAQ content into seed data, and update the FAQ page to read from the DB. Same shape of change as #203, could ship as one PR covering both if scoped together.

### Dependencies
None required, but most useful once an admin CRUD section (#202) exists to edit the data through.

david-fs-valente:
## Summary

Implements #204: FAQ content moves from hardcoded `src/edition/FAQ.ts` into the database, editable through the admin backoffice without a deploy.

- **`FaqEntry` model** (`question`, `answer`, `order`) + migration. Since `prisma/seed.ts` is disabled in production, the migration itself backfills the current 12 FAQ entries (same pattern as `add_sponsor_table`'s backfill) — production gets the existing content automatically on `migrate:deploy`, no manual re-entry needed.
- **Admin CRUD** at `/faqs` (list/create/edit/delete), following the existing `DataTable`/`AdminForm` pattern used by Sponsors/Interests.
- **Drag-and-drop reorder board** at `/faqs/order`, built on the same `@dnd-kit` setup as the Company tier board — single lane (just one ordered list), explicit "Guardar" commits the new order in one bulk transaction, never autosaves on drag. No validation needed here (unlike the timetable in #203, this is pure ordering, not time-based).
- **Public consumption**: new `GET /api/faqs` route; `FaqSection` now fetches client-side (same `useEffect` + `httpClient.get` pattern as `SponsorsSection`/`CompaniesSection` — it sits inside a `"use client"` page tree, so it can't become an async Server Component) instead of importing the now-deleted `edition/FAQ.ts`.
- **`AdminForm` gains a `"textarea"` field kind** — the existing `"text"`/`"email"`/`"number"` kinds are all single-line, a poor fit for paragraph-length FAQ answers.
- **AGENTS.md updated**: FAQ (and Sponsors, already DB-backed before this PR) dropped from the `edition/`-owned content description, the "Edition-specific content" convention paragraph, and the "where new code goes" table.

## Verification

- `pnpm typecheck` — clean.
- `pnpm test` — 120/120 passing (35 files), including a new `AdminForm` test for the `textarea` kind.
- Verified the full migration chain applies cleanly from scratch against a real Postgres instance, and that all 12 FAQ rows land with correct ordering and Portuguese diacritics intact.
- `pnpm dev` against that same local Postgres: `GET /api/faqs` → 200 with correct JSON; `GET /api/admin/faqs` and the `/faqs`, `/faqs/new`, `/faqs/order` pages → 401/404 as expected without a session (no 500s); homepage → 200.
- **Not verified**: the actual admin UI (list/form/drag-and-drop) end-to-end, since no real Supabase project was available in this environment to authenticate as an admin — code review + the underlying route/page auth checks were verified instead. Please click through `/faqs` once deployed to a real environment before considering this fully done.

Refs #204


david-fs-valente:
## Blocking

1. **New FAQ entries default to `order: 0`, jumping to the top of the public list.** `FaqForm` (`src/components/FaqForm/index.tsx`) never sends an `order` value, so every entry created via "Adicionar pergunta" lands at the DB default (`order: 0`) — ahead of all existing entries — until an admin manually visits `/faqs/order` to fix it. This happens on every normal use of the "add" flow, not just an edge case. Suggest defaulting new entries to `max(order) + 1` server-side in `createFaqEntryForAdmin` (`src/application/services/faqService.ts`) instead of relying on the Prisma column default.

david-fs-valente:
## Non-blocking

1. **Duplicate `question` surfaces as a generic 500.** `question` is `@unique` in the `FaqEntry` model, but `createFaqEntryForAdmin`/`updateFaqEntryForAdmin` (`src/application/services/faqService.ts`) never catch the resulting Prisma `P2002` — it falls through `defineHandler`'s catch-all (`src/lib/http/server.ts`) to `{ error: "Something went wrong" }`/500 instead of a friendly 400.
2. **`updateFaqOrder` doesn't verify ids still exist before the bulk transaction**, unlike `updateScheduleOrder` in #248. A stale/deleted id (e.g. deleted in another tab while the order board is open) surfaces as an unhandled Prisma `P2025` inside `bulkUpdateFaqOrder`'s `$transaction` → generic 500 rather than a clean 404.
3. **No `faqService`/`faqRepository` unit tests** — only an `AdminForm` textarea-kind test was added. #248 added solid coverage for the equivalent schedule logic (`scheduleService.test.ts`); worth mirroring here for parity (e.g. the "not found" 404 branches, the order bulk-update path).
4. **FYI — merge-conflict heads-up:** `AGENTS.md`, `src/edition/index.ts`, and `src/components/AdminForm/index.tsx` are all also touched by #248. Expect a real conflict in `edition/index.ts` (one PR removes the `FAQ` import/export, the other removes `ScheduleDays`) and the AGENTS.md prose paragraph when merging whichever PR lands second. The `AdminForm` union-type reformat is identical in both, so that hunk should auto-resolve.

david-fs-valente:
## Non-blocking issues

- **No route-level tests for the new FAQ endpoints.** `faqService.ts` has good unit coverage, but `/api/faqs`, `/api/admin/faqs`, `/api/admin/faqs/[id]`, and `/api/admin/faqs/order` have no tests exercising the actual `defineHandler` wiring (auth Strategy, schema parsing, status codes). PR #249 establishes exactly this pattern for other routes (e.g. `actions/[id]/route.test.ts`, `qrcode/route.test.ts`) — worth following here too.
- `createFaqEntryForAdmin`'s `findMaxFaqOrder()` read and the subsequent `createFaqEntry()` write aren't transactional, so two concurrent admin creates could compute the same "next order" value. Low risk given this is a single-admin-at-a-time backoffice tool and `order` has no DB uniqueness constraint, but worth knowing.
- The `AdminFormSection` union-type reformat in `src/components/AdminForm/index.tsx` (removing the leading-`|` style) touches the exact same lines that PR #248 also modifies independently — expect a merge conflict whichever of the two lands second.

No blocking issues found — the FAQ→DB migration, admin CRUD, and reorder board all look correct.

david-fs-valente:
### Code review: non-blocking notes

No blocking issues found. A few things worth a look, none of which need to hold up merge:

1. **`FaqSection` regresses public FAQ content from SSR to a client-side fetch** (`src/components/Faq/FaqSection/index.tsx`) — the section now renders `null` until a `useEffect` fetch resolves, so the FAQ list is no longer in the initial server-rendered HTML (SEO/no-JS impact, plus a load flash). This mirrors the existing `SponsorsSection` pattern, so it's consistent with prior art rather than a new mistake — flagging in case the SSR tradeoff wasn't a deliberate call for this content specifically.
2. **Silent fetch failure**: `.catch(() => setFaqs([]))` swallows any fetch error with no visible error state or logging.
3. **`updateFaqOrder` fetches the entire FAQ table just to validate the submitted ids exist** (`faqService.ts`) — fine at current scale (~12 rows), but a `findMany({ where: { id: { in } } })` would scale better if this pattern gets reused elsewhere.
4. **Minor typing nit**: `ADMIN_SORTABLE_FIELDS.includes(sort as AdminFaqSortField)` casts before the membership check, which defeats the point of the check for TS (works fine at runtime).
5. **`FaqBoard` only wires up `PointerSensor`** — drag-to-reorder is mouse/touch only, no keyboard-accessible fallback for admins.
6. **No component-level tests** for the new `FaqBoard`/`FaqForm` client components — coverage stops at the route/service layer (acknowledged in the PR's own "not verified" note).

david-fs-valente:
Pushed two more fixes on top of the earlier round:

- `FaqForm` was still swallowing the real server error behind a generic toast (e.g. the 409 for a duplicate question never reached the admin) - now surfaces `HttpClientError.message` when available, mirroring the `ScheduleForm` fix in #248. Added `FaqForm/index.test.tsx` (no component-level test existed before).
- `updateFaqOrder`'s id-existence check now uses a targeted `findFaqEntriesByIds` query instead of fetching the whole table.

Left a few of the earlier non-blocking notes alone on purpose, rather than fixing them one-off in this PR:

- **SSR→client-fetch + silent `.catch(() => setFaqs([]))` in `FaqSection`**: identical to the existing `SponsorsSection` pattern. Fixing it only for FAQ would diverge from that prior art rather than resolve something this PR introduced.
- **`ADMIN_SORTABLE_FIELDS.includes(sort as X)` cast-before-check nit**: the same shape is already used in `sponsorRepository`, `employeeRepository`, `actionRepository`, `studentRepository`, `interestRepository`, and `companyRepository`. Changing it just here would be inconsistent with the rest of the admin-repository layer - a repo-wide pass would be the right scope for that, not this PR.
- **`FaqBoard`'s `PointerSensor`-only drag-and-drop (no keyboard fallback)**: `CompanyTierBoard`, the board this was built from, has the same gap. Same reasoning - a real accessibility fix belongs across all three boards at once, not bolted onto one.

Everything else from the review (order defaulting to 0, duplicate-question 500s, the reorder id-existence check, service/route test coverage) was already fixed in the prior round of commits. `pnpm lint`/`typecheck`/`test` are clean (148/148 passing).

PR #248, opened 2026-07-23
feat: move timetable/schedule data into the database

### LINKED ISSUE #203: [Refactor] Move timetable/schedule data into the database

david-fs-valente:
## Summary

Implements #203: the timetable moves from hardcoded `src/edition/ScheduleDays.ts` into the database, editable through the admin backoffice, with a drag-and-drop reorder board that enforces chronological consistency.

- **`ScheduleEvent` model** (`day`, `order`, `startTime`, `endTime`, `activity`) + migration. `ScheduleDays.ts` only ever recorded a single "hour" per entry — this PR adds a real `endTime` per row, seeded as the next entry's start within the same day (a 30-minute placeholder for each day's last entry, editable afterward). Since `prisma/seed.ts` is disabled in production, the migration itself backfills the current 16 entries (same pattern as `add_sponsor_table`/`add_faq_entry`).
- **Admin CRUD** at `/schedule` (list/create/edit/delete).
- **Two-lane drag-and-drop reorder board** at `/schedule/order` — one lane per day (matching the two-day assumption already baked into `Schedule`/`Content`), same `@dnd-kit` setup as the Company tier board and the FAQ board (#204).
- **The validation you asked for**: `domain/schedule/scheduleValidation.ts` is a pure function that flags any row starting before the previous row (in display order) has ended, or whose own end isn't after its own start. The board recomputes this on every drag and gives the offending row's start time a red outline; "Guardar" is disabled while any row is invalid. This is re-checked server-side too in `updateScheduleOrder()` before committing — the client-side disable is a convenience, not the actual trust boundary.
- **Public consumption**: new `GET /api/schedule` route; a new `ScheduleSection` fetches client-side (same pattern as `FaqSection`/`SponsorsSection`) and reformats `"09:45"` back to the site's existing `"09.45h"` notation before handing off to the unchanged `Schedule` presentational component — no visual change on the public site.
- **`AdminForm` gains a `"time"` field kind** — reuses the existing generic `<input type={field.kind}>` fallback (native browser time picker), no renderer changes needed.
- **AGENTS.md updated** to reflect the schedule/timetable no longer being `edition/`-owned content.

## Verification

- `pnpm typecheck` — clean.
- `pnpm test` — 131/131 passing (37 files), including:
  - `domain/schedule/scheduleValidation.test.ts` — the pure overlap-check function (consistent sequences, out-of-order start, zero/negative-duration rows, empty/single-row edge cases, back-to-back rows are valid).
  - `scheduleService.test.ts` — `updateScheduleOrder()`'s server-side re-validation (valid reorder commits, an invalid one is rejected without saving, including a cross-day move that would overlap its new neighbour).
- Verified the full migration chain applies cleanly from scratch against a real Postgres instance, with all 16 rows landing with correct day/order/times and Portuguese diacritics intact.
- `pnpm dev` against that same local Postgres: `GET /api/schedule` → 200 with correct JSON; `GET /api/admin/schedule` and the `/schedule/order`, `/schedule/new` pages → 401/404 as expected without a session (no 500s); homepage → 200.
- **Not verified**: the actual admin UI (list/form/drag-and-drop, and specifically the red-outline/disabled-save behavior in a real browser), since no real Supabase project was available in this environment to authenticate as an admin — the underlying validation logic is unit-tested instead. Please click through `/schedule/order` once deployed to a real environment, including dragging a row into an invalid position, before considering this fully done.

Refs #203


david-fs-valente:
## Blocking

1. **Chronological overlap validation only runs on the drag-reorder endpoint, not on plain create/edit.** `domain/schedule/scheduleValidation.ts`'s `findInvalidScheduleRowIds` is only invoked from `updateScheduleOrder`. `createScheduleEventForAdmin`/`updateScheduleEventForAdmin` (used by `/schedule/new` and `/schedule/[id]/edit`) never call it — an admin editing a single event's `startTime`/`endTime` directly (arguably the more common path for changing times, since the drag board only reorders) can introduce an overlap with zero validation, despite the PR description framing server-side re-validation as a general safety net.
2. **`day` isn't constrained to `{1, 2}`.** `createAdminScheduleSchema`/`updateAdminScheduleSchema` (`src/schemas/adminScheduleSchema.ts`) only enforce `z.number().int().positive()`. An out-of-range `day` (e.g. a typo entering `3`) silently disappears from the public schedule, since `ScheduleSection.groupByDay` strictly filters `event.day === 1 | 2` — no error surfaces anywhere except noticing it in the admin list. Suggest tightening to `z.union([z.literal(1), z.literal(2)])`.
3. **New events default to `order: 0`.** `ScheduleForm` never sends `order`, so every "Adicionar atividade" inserts at the top of that day's order — which can immediately trip the very overlap validation this PR adds, forcing an unnecessary trip to `/schedule/order` just to unblock saving. Same underlying issue as `createFaqEntryForAdmin` in #247; suggest defaulting new events to the end of their day's order server-side.

david-fs-valente:
## Non-blocking

1. **`ScheduleBoard`'s day lane labels are hardcoded** (`"Dia 1"`/`"Dia 2"` in `src/components/ScheduleBoard/index.tsx`) rather than sourced from `branding.event.scheduleDayTitles`, which is what the public `ScheduleSection` actually displays. Cosmetic drift risk if those titles ever change for a future edition.
2. **FYI — merge-conflict heads-up:** `AGENTS.md`, `src/edition/index.ts`, and `src/components/AdminForm/index.tsx` are all also touched by #247. Expect a real conflict in `edition/index.ts` and the AGENTS.md prose paragraph when merging whichever PR lands second. The `AdminForm` union-type reformat is identical in both, so that hunk should auto-resolve.

david-fs-valente:
## Blocking issue

**`createScheduleEventForAdmin` rejects legitimate, non-overlapping event creation whenever the new event isn't chronologically last in its day.**

`src/application/services/scheduleService.ts` (`createScheduleEventForAdmin`, ~L758-782): when `order` isn't explicitly supplied — and `ScheduleForm`/`/schedule/new` never supplies it — it defaults to `findMaxScheduleOrderForDay(day) + 1`, i.e. the new row is always appended *last* in the day's `order` sequence. `assertScheduleDayIsValid` then sorts by `order` and checks the new row's `startTime` against the *previous-in-order* row's `endTime`, not against actual chronological neighbors.

Concrete repro: Day 1 has one entry `a` (09:00–10:00). Adding a genuinely non-overlapping earlier activity at 08:00–08:30 via `/schedule/new` gets `order = 1` (after `a`). Validation checks `08:00 < a.endTime (10:00)` → rejected with a 400, even though the two events don't overlap at all in wall-clock time.

In practice, the "Add activity" admin form can only ever successfully create an event that comes *after* every existing event of that day — any "insert earlier / in the middle" case is unconditionally rejected. This undermines the CRUD form for a large class of legitimate edits (e.g. inserting a forgotten session).

This is compounded by `ScheduleForm.handleSubmit`'s catch block always showing a generic `"Não foi possível guardar a atividade."` toast, discarding the actual server error — so there's no indication to the admin of why the save failed or that they'd need to add it at the end and then drag-reorder it into place.

The same root cause (order not derived from time) also affects `updateScheduleEventForAdmin` when an edit changes `day` without an explicit `order` — the row keeps its old day's order value, which may not be a sensible position in the new day.

Suggested fix: when `order` isn't explicitly supplied, derive it by finding where `startTime` actually falls among the day's existing rows (insert at the correct chronological index) rather than defaulting to "append last."

## Non-blocking issues

- No route-level tests for `/api/admin/schedule*` / `/api/schedule` (same gap as PR #247) — only `scheduleService.test.ts` and `scheduleValidation.test.ts` exist.
- `scheduleService.test.ts`'s "rejects creating an event that overlaps an existing row" test actually exercises a variant of the bug above but frames it as intentional overlap-rejection with a genuinely-overlapping input — adding a non-overlapping-but-earlier case would surface this bug directly in the suite.
- The `AdminFormSection` type reformat overlaps with PR #247's identical change on the same lines — expect a merge conflict between the two.

david-fs-valente:
### Code review: non-blocking notes

No blocking issues found. Flagging a few things for consideration, none of which need to hold up merge:

1. **Atomicity gap between the order-shift and the actual write** in `createScheduleEventForAdmin`/`updateScheduleEventForAdmin` (`scheduleService.ts`): when inserting/editing a row requires shifting later rows, `bulkUpdateScheduleOrder` (its own `$transaction`) runs, then `createScheduleEvent`/`updateScheduleEvent` runs as a separate, unrelated write. If the process dies between the two, you can be left with shifted neighbors and either no new row (create) or a row whose own time fields never persisted while its neighbors already moved (update) — a real if narrow data-consistency gap. Worth noting since the sibling FAQ PR (#247) solves the equivalent two-step hazard with `withTransaction`, and `scheduleRepository.ts`'s `createScheduleEvent`/`updateScheduleEvent`/`bulkUpdateScheduleOrder` don't currently accept an optional `db: DbClient` the way `faqRepository.ts`'s do. Recommend wrapping the shift + write in one transaction for consistency with that pattern — reasonable as a fast-follow if not addressed now.
2. **`updateScheduleOrder` validates only the rows included in the request, not the full day** — it never merges in a day's untouched rows still in the DB before running `assertScheduleDayIsValid`. Currently unreachable via the UI (`ScheduleBoard.handleSave` always submits the full board state for both days), but nothing enforces that a caller must send the complete day, so a partial/stale submission could in principle create an order collision the check never sees.
3. Same minor typing nit as #247: `ADMIN_SORTABLE_FIELDS.includes(sort as AdminScheduleSortField)` casts before the membership check.
4. **`ScheduleBoard`** has the same keyboard-accessibility gap as `FaqBoard` in #247 (`PointerSensor` only).
5. **`ScheduleSection`** has the same SSR→client-fetch tradeoff as `FaqSection` in #247.
6. **Heads-up, not a code issue**: this PR and #247 both independently edit `src/edition/index.ts` and the same `AGENTS.md` sections — whichever merges second into `dev` will hit a straightforward merge conflict in both files.

david-fs-valente:
Pushed two more fixes on top of the earlier round:

- Closed the atomicity gap between the order-shift and the actual write: `createScheduleEventForAdmin`/`updateScheduleEventForAdmin` ran `bulkUpdateScheduleOrder` as its own transaction and then `createScheduleEvent`/`updateScheduleEvent` as a separate write - a process dying in between could leave shifted neighbours with no corresponding new/updated row. `createScheduleEvent`, `updateScheduleEvent`, and `bulkUpdateScheduleOrder` now take an optional db client, and both service functions wrap the shift + write in one `withTransaction`, matching the pattern `faqService.createFaqEntryForAdmin` already uses (#247).
- `updateScheduleOrder` only validated the rows present in the submitted `updates` array, so a caller that didn't send a day's full row set could get a collision with an untouched row past validation - currently unreachable via the UI (the board always submits the full state) but nothing enforced that. It now builds each affected day from every row in the DB, applying the submitted day/order where present and falling back to the row's current values otherwise. Added a regression test for the partial-submission case.

Left a few of the earlier non-blocking notes alone on purpose, rather than fixing them one-off in this PR:

- **SSR→client-fetch + silent `.catch(() => setEvents([]))` in `ScheduleSection`**: identical to the existing `SponsorsSection`/`FaqSection` pattern - fixing it only here would diverge from that prior art rather than resolve something this PR introduced.
- **`ADMIN_SORTABLE_FIELDS.includes(sort as X)` cast-before-check nit**: same shape already used in `sponsorRepository`, `employeeRepository`, `actionRepository`, `studentRepository`, `interestRepository`, and `companyRepository` - a repo-wide pass is the right scope for that, not this PR.
- **`ScheduleBoard`'s `PointerSensor`-only drag-and-drop (no keyboard fallback)**: `CompanyTierBoard`, the board this was built from, has the same gap - same reasoning, belongs across all three boards at once.

Everything else from the review (overlap validation on plain create/edit, `day` constrained to 1|2, order defaulting to 0, the append-last insertion bug, hardcoded day labels, service/route test coverage, the generic error toast) was already fixed in the prior round of commits. `pnpm lint`/`typecheck`/`test` are clean (152/152 passing).

david-fs-valente:
Addressed the concurrency-gap note directly:

> Real concurrency gap in the "server is the trust boundary" claim: createScheduleEventForAdmin, updateScheduleEventForAdmin, and updateScheduleOrder all call findAllScheduleEvents() outside any transaction to build the day's rows, validate, then write inside withTransaction/bulkUpdateScheduleOrder.

Fair - the earlier "atomicity" commit only wrapped the *write* half (shift + create/update) in a transaction; the read that the plan and validation are computed from still ran before the transaction opened, so a concurrent request could invalidate that snapshot before it committed.

Fixed by giving `findAllScheduleEvents`/`findScheduleEventById` an optional db client and moving the read (plus the plan/validation that depends on it) inside the same `withTransaction` callback as the write, in all three functions. That closes the window down to the transaction's own duration instead of a full read-then-decide-then-write round trip through the application layer.

To be upfront about the actual guarantee this gives: it's the "re-check inside the transaction" mitigation you suggested, not a full fix. Postgres's default read-committed isolation doesn't lock rows on a plain `SELECT`, so two transactions that both start their read before either commits could still theoretically race. Closing that fully would need row locking (`SELECT ... FOR UPDATE`), serializable isolation with retry-on-conflict, or a DB constraint enforcing (day, order)/non-overlap invariants - all bigger changes than fit this PR. Logged as a follow-up todo (covering the same class of issue in the FAQ ordering code too, even though FAQ's create already reads its max-order inside its transaction).

`pnpm lint`/`typecheck`/`test` clean (152/152 passing).

david-fs-valente:
#247 merged into \`dev\` first, so this PR picked up the anticipated conflicts. Merged \`dev\` into this branch and resolved:

- **\`prisma/schema.prisma\`**: kept both new models (\`ScheduleEvent\` from this PR, \`FaqEntry\` from #247) side by side. Both migrations already live in separate timestamped folders, so no migration-level conflict.
- **\`src/edition/index.ts\`**: dropped both the \`FAQ\` and \`ScheduleDays\` imports/exports - each PR only removed its own, so the merge needed both gone together.
- **\`AGENTS.md\`**: reworded the three conflicting spots (directory-tree comment, the "Edition-specific content" bullet, and the "where new code goes" table row) to reflect that Sponsors, FAQ, *and* the schedule/timetable are all DB-backed now - only tier company lists, branding, and \`actions.ts\` are still hardcoded in \`edition/\`.
- **\`src/components/AdminSidebar/index.tsx\`**: kept both new nav items ("Schedule" and "FAQs") under the Edition section instead of picking one.
- **\`src/components/AdminForm/index.tsx\`** auto-merged cleanly, as expected - it now has both the \`"time"\` and \`"textarea"\` field kinds.

Regenerated the Prisma client against the merged schema and reran the full suite: `pnpm lint`/`typecheck` clean, `pnpm test` 207/207 passing (53 files, combining both PRs' test suites). Pushed as a merge commit (`f239e32`); GitHub now reports this branch as mergeable into `dev`.

Issue #222, opened 2026-07-19
[Chore] Verify Supabase storage bucket privacy config for avatars/cvs


david-fs-valente:
### Description
No bucket-creation script or migration exists in-repo for the `avatars` (public) / `cvs` (private) Supabase Storage buckets — they're only referenced by name in application code (`utils/` Supabase client factories, `exportService.ts`, the upload routes). Surfaced by the #207 security audit.

### Why
Static analysis can't confirm from the codebase alone that the `cvs` bucket is actually configured private in the Supabase project (vs. relying on dashboard convention that could drift or be misconfigured on a new environment/edition setup). This is exactly the kind of config-drift static review can't catch — worth a direct one-time check in the Supabase dashboard, and worth capturing however that config is verified going forward (e.g. an infra-as-code bucket definition, or a documented manual step in README's setup instructions) so a future environment doesn't silently create `cvs` as public.

### How
Confirm current bucket ACLs directly in the Supabase dashboard for the production project. If satisfactory, document the expected config (README or a setup script) so it's reproducible for future environments/editions rather than tribal knowledge.

### Dependencies
None.

PR #249, opened 2026-07-23
test: expand coverage per #207 audit + deploy-config smoke test

### LINKED ISSUE #209: [Test] Expand test coverage (integration + deploy-config smoke tests)
### Description
Current test coverage (per AGENTS.md) covers application boundaries/services, domain value objects, edition action rules, auth flow, saved-student comments, logger/Sentry privacy, ISEP email normalization, and component smoke tests — but gaps remain, especially around integration paths. The recent `NEXT_PUBLIC_BASE_URL`-missing-from-Docker-build-args bug is a concrete example of a class of bug unit tests alone wouldn't catch.

### Why
Unit tests catch logic bugs; they don't catch wiring/config/deploy-shape bugs, which is exactly the category that's bitten this project recently. Closing that gap reduces the chance of another silent prod-only failure.

### How
Add: more unit tests where coverage is thin, smoke tests for deploy-sensitive config (env var / build-arg wiring), and integration tests for multi-step flows (auth, storage upload, QR scan). Scope the specific list after #207 (security audit) surfaces which paths matter most — don't front-load test-writing before knowing what the audit finds.

### Dependencies
Depends on #207 (security audit) — scope test additions from what that surfaces.

david-fs-valente:
## Summary

Implements #209, scoped now that #207's security audit is fully resolved (all 8 follow-ups — #215–#222 — closed). Rather than front-loading generic coverage, this targets: (1) a real, currently-live bug in the exact class #209 was written about, (2) the audit-derived paths that came back with zero or partial coverage, and (3) the multi-step auth flows #209's own "How" section calls out (QR scan) that had none at all.

### Fixed a live deploy-config bug + added the smoke test that would've caught it

`NEXT_PUBLIC_LOGS_DASHBOARD_URL` (added for the admin backoffice's GlitchTip link, #241) was declared in `env.client.ts` but never wired into the Dockerfile's `app-builder` stage or `docker-compose.app.yml`'s build args — the exact same silent-failure shape as the `NEXT_PUBLIC_BASE_URL` incident that motivated this issue, recurring for a different var. Since Next.js inlines `NEXT_PUBLIC_*` vars at build time, this meant the Logs link would silently be blank in any Docker-built production deploy, however correctly the var was set in the environment.

- Fixed both the Dockerfile and `docker-compose.app.yml`.
- Added `tests/e2e/dockerBuildArgs.test.ts`, which closes the whole class of bug rather than just this instance: it reads every `NEXT_PUBLIC_*` var declared in `env.client.ts` and asserts each one is wired into both files. Verified it actually fails without the fix and passes with it before committing either.

### QR/action-scan flow — previously zero coverage

- `authService.test.ts`: the real `signJwt`/`verifyJwt` (every existing consumer mocked `authService` away, so the actual JWT sign/verify logic itself was never exercised) — round-trip, rejects garbage/foreign-secret/expired tokens.
- `qrcode/route.test.ts` and `actions/[id]/route.test.ts`: auth gating, the scanned-token verification path, and the admin-only `PATCH` toggle. Includes a regression case for #212 (the ~8.3-hour QR token expiry bug): signs a token with the real 30s expiry, advances fake timers past it, confirms the route now correctly rejects it instead of completing the action.
- Found + fixed a **stale AGENTS.md claim** while writing these: the "~8.3 hour token" and "anti-replay check commented out" gotchas predate #212's fix and no longer matched the code.

### Student profile ownership check (#220) — previously zero coverage

`page.test.tsx` for `(profiles)/student/[...data]` covers the exact comparison #220 fixed (`.match()` → `!==`): own profile allowed, a different student blocked, an employee whose company saved the student allowed, one that didn't blocked, no session blocked. Needed a `"~"` alias added to `vitest.config.ts` (mirrors `tsconfig.json`'s existing mapping) since Vitest didn't know how to resolve the page's `~/public/*.png` static asset import at all — a first for this codebase's tests.

### `/api/students/[code]/stats` (#216) — partial coverage

Added the missing "company that saved the student" case to `savedStudentService.test.ts` (only "own stats" and "unrelated student rejected" existed), plus the first route-level test for the actual HTTP handler (auth gating, correct forwarding of ownership info, 404 propagation) — previously only the service logic was tested, not the wiring.

## Not included (lower value, flagging rather than doing)

- A CSPRNG-usage assertion for `GenerateCode.ts` (#217) — existing tests already cover format/uniqueness; the fix itself already uses `crypto.randomInt`.
- A full authenticated-success-path test for the avatar/CV upload routes (#215) — existing tests already prove the auth check passes, just stop at the "missing file" 400 rather than a full successful upload.

## Verification

- `pnpm typecheck` — clean.
- `pnpm test` — 145/145 passing (41 files, up from 35).
- Manually confirmed `dockerBuildArgs.test.ts` fails without the Dockerfile/compose fix and passes with it (stashed and restored the fix to verify both directions before committing).

Refs #209


david-fs-valente:
## Non-blocking

1. `extractComposeWebBuildArgs()` in `tests/e2e/dockerBuildArgs.test.ts` locates the *first* `args:` block in the whole compose file rather than scoping specifically to the `web` service. That's safe today (`migrate` has no `args:` block), but if a future PR gives `migrate` its own build args positioned before `web` in the file, this would silently validate the wrong block — a quiet false-negative in the exact test meant to prevent silent config gaps like the one this PR fixes. Consider anchoring the search to `services:\n  web:` first.
2. `readClientEnvVarNames()`'s regex scan over `env.client.ts`'s raw source text will also match `NEXT_PUBLIC_*` names that appear only in comments, not just schema keys. Not a problem today, but worth remembering if a future comment references a var name that isn't actually still declared — it'd demand that name be wired into Docker even though it's not real.

Both are minor robustness notes on brand-new test infrastructure, not blockers — verified the Dockerfile/compose fix and the new tests' claims directly against the PR branch and they're all accurate.

david-fs-valente:
## Non-blocking issues

- `isBlocked()` in `src/app/(profiles)/student/[...data]/page.test.tsx` (`element.type === "section"`) infers "access blocked" from `Custom404`'s current return shape (a bare `<section>`) rather than a more structurally stable signal (e.g. a test id, or the component reference itself). The in-file comment explains the reasoning, which helps, but a future refactor of `not-found.tsx`'s return shape would silently break this test's assumption rather than fail loudly.
- `students/[code]/stats/route.test.ts`'s second test asserts `getStudentStats` was called with `{ studentCode: undefined, companyId: "c1", isAdmin: false }` — asserting on the incidental `studentCode: undefined` field couples the test to the exact call-object shape rather than just the behavior under test. Not wrong, just a bit brittle.

No blocking issues found — the Dockerfile/compose fix is real and well-verified (confirmed to fail without it, pass with it), and the new auth/QR/ownership test coverage is solid and correctly targeted at the previously-uncovered paths.

PR #268, opened 2026-07-28
feat: add SuperAdmin/Admin role tiers

### LINKED ISSUE #269: [Feat] Add SuperAdmin/Admin role tiers
### Description
Replace the flat `User.isAdmin` boolean with two admin tiers: **Admin** (everything today's admin backoffice already covers - Students, Employees, Companies, Sponsors, Actions, Interests, Schedule, FAQs, Storage, Statistics) and **Super Admin** (everything Admin can do, plus managing admin accounts themselves - creating, editing, activating/deactivating).

### Why
Right now every admin account has identical, unrestricted access to the entire backoffice, including the ability to (indirectly, via direct DB/Supabase access) grant admin rights to anyone - there's no way to have a smaller circle of trusted people control who else gets admin access versus a wider group that just needs day-to-day Student/Employee management. Splitting this into two tiers lets the org delegate routine admin work broadly while keeping account-granting power narrow.

### How
- `User.isAdmin: Boolean` → `User.adminRole: AdminRole?` (`ADMIN | SUPER_ADMIN`, null = not an admin). Migration backfills existing `isAdmin=true` rows to `adminRole=ADMIN` - nobody is auto-promoted to Super Admin.
- `User.role` becomes nullable - an admin-only account has no Student/Employee profile, so `role` is genuinely `null` rather than an arbitrary default. New `User.name` holds an admin account's display name (Student/Employee already have their own on their profile table).
- `authPolicy.ts` gets a new `"superadmin"` policy (`adminRole === "SUPER_ADMIN"`) alongside the existing `"admin"` one (`adminRole !== null`, i.e. either tier - same meaning the old `isAdmin` check had).
- New `/admins` section in the admin backoffice, visible and reachable only for Super Admins (sidebar entry + page-level guard, on top of the API-level `"superadmin"` policy): create/edit admin accounts (mirrors the existing Employee admin-CRUD pattern - `admin.auth.admin.createUser` with the password set directly, display name passed through to Supabase's `user_metadata`), activate/deactivate, delete. A super admin can't deactivate or delete their own account through the panel (would risk locking every admin out), but can demote themselves.

### Dependencies
None.

david-fs-valente:
## Summary

Closes #269.

Replaces the flat `User.isAdmin` boolean with two admin tiers:
- **Admin** — everything today's admin backoffice already does (Students, Employees, Companies, Sponsors, Actions, Interests, Schedule, FAQs, Storage, Statistics).
- **Super Admin** — everything Admin can do, plus a new Admins section to create/edit/activate/deactivate admin accounts.

## Schema
- `User.isAdmin: Boolean` → `User.adminRole: AdminRole?` (`ADMIN | SUPER_ADMIN`, null = not an admin).
- `User.role` becomes nullable — an admin-only account has no Student/Employee profile, so `role` is genuinely `null` rather than an arbitrary default.
- New `User.name: String?` for an admin account's display name (Student/Employee already have their own `name` on their profile table).
- Migration backfills existing `isAdmin=true` rows to `adminRole=ADMIN` before dropping the old column — **nobody is auto-promoted to Super Admin by this migration.**

## Auth
- `authPolicy.ts` gets a new `"superadmin"` policy (`adminRole === "SUPER_ADMIN"`) alongside the existing `"admin"` one (`adminRole !== null` — either tier, same meaning as the old `isAdmin` check).
- Every existing `session.isAdmin` read across the app (student-visibility checks, giveaway picker, admin storage upload, saved-student dedup, the admin layout gate) now reads `session.adminRole !== null` — same behavior, just the new field.
- `domain/student/studentAccess.ts`'s own `StudentAccess.isAdmin` is untouched — it's an app-level domain concept derived from `adminRole` at each call site, not the DB field itself.

## New: Admin account management (Super Admin only)
- `POST/GET /api/admin/admins`, `PATCH/DELETE /api/admin/admins/[id]` — gated by the new `"superadmin"` policy.
- New `/admins` section in the admin backoffice (only visible in the sidebar, and only reachable, for Super Admins — page-level `notFound()` guard as defense-in-depth alongside the API gate).
- Creation mirrors the existing Employee admin-CRUD pattern: `admin.auth.admin.createUser` with the password set directly (no invite flow), display name passed through to Supabase's `user_metadata`.
- **Safety guard:** a super admin can't deactivate or delete their own account through this panel (would risk locking every admin out) — demoting themselves is still allowed, since that's recoverable by any other super admin.

## Test plan
- [x] `pnpm lint` on all touched/new files
- [x] `pnpm typecheck`
- [x] `pnpm test` — `authPolicy` (new superadmin cases), `boundaries`, `authApplicationService`, the student profile page, every admin-gated route test, `qrcode`, `actions/[id]`, `serviceLogic`, `savedStudentService`, and the new `adminAccountService` suite (create-rollback, self-deactivate/delete guard, self-demote allowed) all green
- [x] Verified the migration against a throwaway local Postgres: seeded a pre-migration `isAdmin=true` row, applied the migration, confirmed it backfilled to `adminRole='ADMIN'` correctly and `prisma migrate diff` shows no drift afterward
- [ ] Live click-through of the new Admins UI — not done in this environment (no real Supabase session available), but every code path is covered by the tests above

### DIFF
--- a/prisma/migrations/20260728170000_add_admin_roles/migration.sql
+++ b/prisma/migrations/20260728170000_add_admin_roles/migration.sql
@@ -1,0 +1,15 @@
+-- CreateEnum
+CREATE TYPE "AdminRole" AS ENUM ('ADMIN', 'SUPER_ADMIN');
+
+-- AlterTable: add the new columns first, backfill from the boolean being
+-- replaced, then drop it - existing isAdmin=true users become ADMIN (the
+-- narrower tier; nobody is auto-promoted to SUPER_ADMIN by this migration).
+ALTER TABLE "User"
+  ADD COLUMN     "adminRole" "AdminRole",
+  ADD COLUMN     "name" TEXT,
+  ALTER COLUMN "role" DROP NOT NULL,
+  ALTER COLUMN "role" DROP DEFAULT;
+
+UPDATE "User" SET "adminRole" = 'ADMIN' WHERE "isAdmin" = true;
+
+ALTER TABLE "User" DROP COLUMN "isAdmin";
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -9,15 +9,24 @@ datasource db {
 }
 
 model User {
-  id      String  @id @default(uuid()) @db.Uuid
-  email   String  @unique
-  role    Role    @default(STUDENT)
-  isAdmin Boolean @default(false)
+  id        String     @id @default(uuid()) @db.Uuid
+  // Null for an admin-only account (no Student/Employee profile - see
+  // adminRole below). Non-null for every self-service/admin-provisioned
+  // Student or Employee account.
+  role      Role?
+  email     String     @unique
+  // Null means not an admin. Orthogonal to role: an admin account has no
+  // Student/Employee profile and role is null, it isn't a STUDENT/EMPLOYEE
+  // that also happens to be an admin.
+  adminRole AdminRole?
+  // Display name for an admin-only account - Student/Employee have their
+  // own `name` on their respective profile table instead.
+  name      String?
   // Gates a Student/Employee's ability to log in - deactivating the User
   // deactivates both, since Student/Employee are 1:1 with User on this id.
-  active    Boolean  @default(true)
-  createdAt DateTime @default(now())
-  updatedAt DateTime @updatedAt
+  active    Boolean    @default(true)
+  createdAt DateTime   @default(now())
+  updatedAt DateTime   @updatedAt
 
   student   Student?
   employee  Employee?
@@ -181,6 +190,14 @@ model SavedStudent {
 enum Role {
   STUDENT
   EMPLOYEE
+}
+
+enum AdminRole {
+  // Manages Students/Employees and the rest of today's admin backoffice.
+  ADMIN
+  // Everything ADMIN can do, plus creating/editing/deactivating other admin
+  // accounts.
+  SUPER_ADMIN
 }
 
 enum Tier {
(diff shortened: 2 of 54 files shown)


david-fs-valente:
**Blocking: sole Super Admin can demote themselves out of existence**

`updateAdminAccount()`'s `assertNotActingOnSelf()` (`src/application/services/adminAccountService.ts`) only fires when `active === false`, so it blocks self-deactivation and self-deletion but **not** self-demotion (`SUPER_ADMIN` → `ADMIN`). The migration explicitly promotes nobody to `SUPER_ADMIN`, and `seedAdmin()` only creates one. If that one super admin demotes themselves — e.g. by mistake, before a second super admin exists — `/admins` becomes unreachable (gated by both the page-level `notFound()` check and the `"superadmin"` auth policy), with no UI path back. Recovery requires a direct DB write.

The PR description frames self-demotion as safe because it's "recoverable by any other super admin," but that assumption doesn't hold in the single-super-admin state this migration bootstraps into. Suggest extending the guard to reject an action (demote/deactivate/delete) that would leave zero `SUPER_ADMIN`s, rather than just blocking actions on your own account.

david-fs-valente:
**Non-blocking: admin email schema bypasses the repo's `EmailSchema` convention**

`createAdminAccountSchema` (`src/schemas/adminAccountSchema.ts`) validates `email` with raw `z.email()` instead of the `EmailSchema` used by every other signup schema in the repo (e.g. `employeeSignUpSchema.ts`). `createAdminAccount()` then separately calls `Email.create(input.email)` in the service to get the branded `Email` type, so validation effectively runs twice — this is the only signup path that doesn't hand the schema-parsed body straight through as an `Email`. Not a bug, just inconsistent with the established pattern; worth switching to `EmailSchema` for consistency.

Issue #11, opened 2025-10-02
[Task] Update Profile Scan Section of Profile Page (Company)


david-fs-valente:
Update the Profile Scan section of the Profile page design according to the material produced by the Image Department.

PR #271, opened 2026-07-28
feat: add admin overview dashboard

### LINKED ISSUE #202: [Feat] Add a full admin CRUD section (users, data, storage)

david-fs-valente:
## Summary

Follow-up to #270 (which itself stacks on #268/#269) — a real "backoffice template" landing page for admins, rather than dropping them on the Students list. Stacked on #270 since it's a direct continuation of the redirect-target discussion there.

- **New `/overview` page** (any admin tier, gated the same way Statistics/Students already are - the shared admin layout, no extra per-page check needed): five stat tiles (students, companies, recruiters, action completions, saved students), a weekly activity bar chart, and a merged recent-activity feed.
- **`getAdminDashboardSummary()`**: fetches the 5 counts + last-10 recent events from each of 3 sources (student signups, QR action completions, company saves) + last-7-days raw timestamps, in parallel. Merges the 3 recent-event sources into one newest-first feed (capped at 15). Buckets the 7-day window into daily counts per category in JS rather than a DB-side `date_trunc` — this is a single-event fair, not a high-volume product, so the row counts involved stay small enough that this is simpler and more portable.
- **Chart colors**: used the design system's validated categorical palette (dataviz skill) — slots 1-3 (blue/orange/aqua), the only three that clear the CVD and normal-vision floors under all-pairs comparison in both light/dark, which matches exactly this 3-series case. Ran the palette validator before using it rather than eyeballing.
- **No dark mode** on the new page — matches the rest of the admin backoffice (Sidebar, DataTable, Statistics, etc.), none of which has any dark-mode handling today. Introducing it for just this one page would be inconsistent, not more correct.
- **Chart doesn't label every bar** — that's the exact anti-pattern the dataviz skill flags. Exact figures live in a collapsible `<details><table>` instead (visible to sighted users, not just screen readers — an `sr-only` table wouldn't actually help the low-contrast-series relief requirement, since that's a low-vision concern, not a blindness one).
- **Simplification I made deliberately**: skipped a JS hover/tooltip layer on the chart (the skill's default recommendation for any chart with a plot). The collapsible table already gives exact numbers, and this is a low-traffic internal tool, not a public product — didn't think a client component + tooltip state justified itself here. Worth reconsidering if this page sees heavier real-world use than expected.
- `AdminSidebar` gets a new top-level "Dashboard" link to `/overview`, above the existing sectioned nav. `LoginPage` and `UserButton` (both from #270) now redirect admins here instead of `/students`.

## Test plan
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` — new `adminDashboardService` suite (stat passthrough, feed merge/sort/cap, day-bucketing including zero-event days) + updated `LoginPage`/`UserButton` redirect assertions, all green
- [x] Verified the SVG bar-path geometry standalone (zero-height, tiny, max-height, narrow-width-radius-clamp cases) — no NaN/negative/degenerate output
- [ ] Live click-through as an admin — not done in this environment (no real Supabase session available), same limitation as the rest of this feature chain

### DIFF
--- a/src/app/(admin)/overview/page.tsx
+++ b/src/app/(admin)/overview/page.tsx
@@ -1,0 +1,34 @@
+import AdminActivityFeed from "@/components/AdminActivityFeed";
+import AdminStatTile from "@/components/AdminStatTile";
+import AdminWeeklyActivityChart from "@/components/AdminWeeklyActivityChart";
+import { getAdminDashboardSummary } from "@/application/services/adminDashboardService";
+
+const OverviewPage = async () => {
+  const summary = await getAdminDashboardSummary();
+
+  const tiles = [
+    { label: "Estudantes", value: summary.studentCount },
+    { label: "Empresas", value: summary.companyCount },
+    { label: "Recrutadores", value: summary.employeeCount },
+    { label: "Ações completadas", value: summary.actionCompletionCount },
+    { label: "Estudantes guardados", value: summary.savedStudentCount },
+  ];
+
+  return (
+    <section className="flex flex-col gap-6 p-8">
+      <h1 className="text-2xl font-bold text-gray-800">Dashboard</h1>
+
+      <div className="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-5">
+        {tiles.map((tile) => (
+          <AdminStatTile key={tile.label} {...tile} />
+        ))}
+      </div>
+
+      <AdminWeeklyActivityChart data={summary.weeklyActivity} />
+
+      <AdminActivityFeed events={summary.recentActivity} />
+    </section>
+  );
+};
+
+export default OverviewPage;
--- a/src/app/(auth)/login/page.test.tsx
+++ b/src/app/(auth)/login/page.test.tsx
@@ -93,7 +93,7 @@ test("sends an admin to the backoffice instead of the homepage, even though role
   render(<LoginPage />);
   submitLogin();
 
-  await waitFor(() => expect(pushMock).toHaveBeenCalledWith("/students"));
+  await waitFor(() => expect(pushMock).toHaveBeenCalledWith("/overview"));
 });
 
 test("sends a super admin to the backoffice too", async () => {
@@ -106,7 +106,7 @@ test("sends a super admin to the backoffice too", async () => {
   render(<LoginPage />);
   submitLogin();
 
-  await waitFor(() => expect(pushMock).toHaveBeenCalledWith("/students"));
+  await waitFor(() => expect(pushMock).toHaveBeenCalledWith("/overview"));
 });
 
 test("falls back to the homepage for a session with no role and no admin tier", async () => {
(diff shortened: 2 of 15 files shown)


david-fs-valente:
**Blocking: recent-activity feed can show stale events instead of the true most-recent 15**

`getAdminDashboardSummary()` (`src/application/services/adminDashboardService.ts`) fetches only the top `RECENT_PER_SOURCE = 10` events from each of the 3 sources (signups, scans, saves) *before* merging them into one feed capped at `ACTIVITY_FEED_LIMIT = 15`.

If any single source produces more than 10 of the true 15 most-recent events across all types — plausible at a career fair, where QR scans can spike well above signups/saves during busy periods — the merge silently drops newer scans (#11-15, never fetched) and instead back-fills the feed with older signups/saves that aren't actually part of the real top 15. The "recent activity" feed can end up both incomplete and out of order relative to what actually happened most recently.

Suggest fetching at least `ACTIVITY_FEED_LIMIT` per source (i.e. bump `RECENT_PER_SOURCE` to 15, or dynamically over-fetch) so the pre-merge truncation can never hide a true top-15 event.

david-fs-valente:
**Non-blocking: day-bucketing mixes local-time and UTC-time date math (+ missing test for the SVG geometry helper)**

`bucketByDay()` and the `since` cutoff in `getAdminDashboardSummary()` build "today"/the window via local-time `Date` mutators (`new Date()`, `.setDate()`, `.setHours()`) but key/compare buckets using `.toISOString().slice(0,10)` (UTC calendar day). This is only self-consistent today because the container has no `TZ` set (defaults to UTC), so local time == UTC time. If a timezone is ever configured for the deploy (plausible for a Portugal-based event app) or the process runs somewhere non-UTC, this would silently produce an off-by-one on the daily buckets near midnight boundaries. Cheap to harden now with `Date.UTC()` / `getUTCDate()` / `setUTCDate()` throughout instead of mixing the two.

Separately: the PR's test plan mentions verifying the SVG bar-path geometry (`roundedTopBarPath` in `AdminWeeklyActivityChart`) for zero-height/tiny/max-height/degenerate cases, but no test file was committed for it (or for any of the 3 new components — only the service layer has tests). Since that verification already happened, capturing it as a small colocated unit test for `roundedTopBarPath` would turn it into a real regression guard instead of one-off manual checking.

PR #276, opened 2026-07-29
fix: harden QR and upload flows

hugo2006alm:
## What changed

- Move student CV/avatar bytes from Next.js to short-lived Supabase signed uploads.
- Rate-limit upload tickets per authenticated student instead of shared event IP.
- Add Storage ticket failure reporting and document required bucket MIME/size limits.
- Use one QR camera stream with rear-camera fallback and decode single-flight protection.
- Refresh personal QR codes before token expiry and enforce action QR refresh-window timestamps.
- Stabilize full validation on Windows.

## Why

Event Wi-Fi NAT could block all users after five uploads, and concurrent server-buffered uploads risked memory pressure. QR camera streams and repeated requests also caused unreliable scanning.

## Validation

- `pnpm lint --quiet`: passed.
- `pnpm typecheck`: passed.
- `pnpm test`: 64 files, 267 tests passed.
- Configured Supabase project checked read-only: `avatars` is public with 5 MiB PNG/JPEG limits; `cvs` is private with 10 MiB PDF limits.


david-fs-valente:
**Blocking: direct-to-storage upload enforcement now lives entirely outside the codebase**

Since avatar/CV upload bytes no longer transit the Next.js server (`src/app/api/storage/avatar/route.ts`, `src/app/api/storage/cv/route.ts`), the server-side `types.includes(contentType)` / `size > maxSize` checks only validate *client-declared* values used to mint the signed-upload ticket. Nothing in the code stops a non-browser caller from requesting a ticket for a small PNG and then `PUT`-ing an arbitrarily large file of any content type straight to the signed URL — the magic-byte check (`matchesDeclaredType`) also moved to run client-side only (`src/client/api/upload.ts`), so the server never verifies actual bytes anymore either.

The README update correctly documents that "bucket restrictions remain the enforcement boundary for direct uploads," but that's a manual, one-time Supabase dashboard step with no CI/test coverage verifying it's actually set. Before merging, please confirm the `avatars` and `cvs` buckets in the target Supabase project(s) actually have the documented MIME-type/size restrictions configured — otherwise this ships with no real server-side enforcement of file type or size for direct uploads.

david-fs-valente:
**Non-blocking notes**

1. **Stale doc comment** — `src/lib/fileSignature.ts`'s comment on `matchesDeclaredType` ("these signatures let us confirm the bytes actually match the declared type") no longer holds for its new call site: it now runs client-side only (`src/client/api/upload.ts`), so it's advisory UX (early rejection for honest clients) rather than something the server can rely on. Worth a comment update so it doesn't read as a server-enforced control.

2. **Possible QR staleness regression** — `ProfileTab.fetchQrcode` (`src/components/QRCode/QRCodeTab/ProfileTab/index.tsx`) now fetches once on mount instead of the old (buggy) every-render effect. That's a correct fix for the render-loop bug, but the personal QR token expires after 30 minutes and there's no periodic refresh anymore, so a student who leaves the tab open past that will see a QR that silently stops scanning. Might be worth a bounded interval refetch (e.g. ~25 min) rather than fetch-once.

Nice, tightly-scoped fixes otherwise — the `Math.round`→`Math.floor` QR-timestamp bucketing fix paired with the new freshness check in `route.ts` is a solid catch, and the per-student rate-limit key correctly targets the NAT false-positive issue instead of just papering over it.

david-fs-valente:
## Review

**What it does:** Switches avatar/CV uploads from server-buffered `multipart/form-data` to short-lived Supabase signed-upload tickets (bytes go browser→Storage directly); moves upload rate-limiting from per-IP to per-authenticated-student (fixes the NAT-blocks-everyone-after-5-uploads bug); adds a second, timestamp-based replay check on action QR codes; and reworks the QR scanner to use `react-zxing`'s built-in camera handling with a rear-camera fallback and single-flight decode guard.

### Strengths
- The `Math.round` → `Math.floor` change in `getActionQrCode` (`actionService.ts`) is a necessary companion to the new `decoded.timestamp > Date.now()` check in the route — `Math.round` could round a timestamp *forward* past "now," which would make the new check reject legitimately-issued tokens. Good catch, correctly reasoned.
- Test rewrites (`avatar/route.test.ts`, `cv/route.test.ts`) correctly prove the fix for the original bug: one student hitting the limit doesn't block a different student.
- `HttpError`/`reportError` usage and explicit status codes follow AGENTS.md conventions.
- Client-side magic-byte check (`isValidFile` in `upload.ts`) is honestly documented as UX-only now that bytes bypass the server.

1. **Security boundary now lives in a manually-run SQL script.** Since bytes go directly to Storage, the *only* real enforcement of file type/size is `supabase/storage-bucket-limits.sql`, which must be run by hand against every environment (including prod). If it's forgotten on any environment, a signed upload URL will accept arbitrary content/size — there's no automated check that it was actually run. Given this repo already has a test (`dockerBuildArgs.test.ts`) specifically to catch "we forgot a manual step across environments," this is the same shape of risk and probably deserves the same treatment eventually.
2. Anonymous requests aren't rate-limited at all on `/storage/avatar` and `/storage/cv` now (limiter key is `session.id`, only checked after auth passes). This matches the existing `defineHandler` pattern repo-wide, so not a regression introduced here specifically — just worth knowing.
3. Minor: an attacker/buggy client can burn their 5/min ticket budget on requests that fail the `types`/`maxSize` check inside the handler (the limiter already counted them). Self-inflicted only, not blocking.

### Test coverage
Good — new tests cover the per-student vs per-IP limiting fix, the ticket-based response shape, and the new QR refresh-window rejection. Confirmed all consumer call sites (`SignUp/FinalStep`, `SignUp/CvStep`, `Profile/ProfileSection`, `Profile/SettingsSection`) only read `.id` off the upload result, so the response-shape change (dropping `contentType`/`size`/`url` for CV) is safe.

david-fs-valente (commented):
## Overview

Moves avatar/CV uploads from server-buffered multipart to short-lived Supabase signed upload URLs (bytes go browser→Storage directly), switches upload rate-limiting from per-IP to per-student, adds a QR-refresh-window check to the action-scan endpoint, fixes a timestamp-bucketing bug, and reworks the QR camera scanner (rear-camera fallback, single-flight decode guard).

## Strengths

- Root-causes the "event Wi-Fi NAT blocks everyone after 5 uploads" bug correctly — keying the rate limiter on `session.id` instead of client IP (`rateLimiter.check(session!.id)`) is the right fix and is directly tested (`"limits one student without blocking another on same event Wi-Fi"`).
- The `Math.round` → `Math.floor` fix in `getActionQrCode` is *load-bearing* for the new expiry check in `actions/[id]/route.ts` (`decoded.timestamp > Date.now()`) — without it, rounding up to a future bucket would make freshly-generated QR codes immediately rejected as "expired." Good catch, and it's covered by a regression test isolating the new check from the JWT's own `exp`.
- Follows house conventions well: new route uses `defineHandler` with `auth: "student"` + a proper `schemas/storageUploadTicketSchema.ts` instead of ad hoc parsing, throws `HttpError` for the 502 path, explicit status codes throughout.
- Transparent about the new trust boundary: README and `fileSignature.ts` comments correctly state that client-side signature/size/type checks are now UX-only, and the real enforcement boundary is the Storage bucket config (`supabase/storage-bucket-limits.sql`).

## Issues / risks

1. **Bucket restrictions are the only real enforcement, but applying them is a manual, unverified step.** `storage-bucket-limits.sql` must be run by hand against *every* Supabase project (staging, prod, and — per this repo's yearly-edition model — every freshly created project for a new event). Nothing in CI, deploy, or the new Playwright suite (#278) actually asserts the bucket rejects an oversized/wrong-type file. If that script is ever skipped for an environment, a signed ticket lets an authenticated student upload an arbitrary file of arbitrary size/type directly to Storage with zero server-side check. Worth a smoke test (e.g., attempt an oversized/mismatched-type upload against a real bucket and expect rejection) rather than relying on doc + memory.
2. **QRCodeScanner (`src/components/QRCode/QRCodeScanner/index.tsx`) drops a documented mobile workaround.** The removed code had this comment: *"on mobile, if the user already gave permissions in the past, [getUserMedia] will still throw an error, even tho the camera opens and works as expected"* — and deliberately swallowed that error so the video kept working. The new code wires `useZxing`'s `onError` (fired when its internal `decodeFromConstraints`/`getUserMedia` call throws) to set a **permanent, blocking** `cameraError` message that replaces the video element entirely. If the flaky-but-actually-working scenario the old comment describes still occurs on real devices, this is a regression — a previously-working scanner would now show a hard error instead of scanning. No new test covers this path. Worth exercising on real Android/iOS devices with previously-granted camera permission before merge, not just `pnpm test`.
3. Minor: rate limiting is presumably in-memory per-process (pre-existing `createRateLimiter`); fine for now but worth remembering if the app ever scales to multiple instances — not introduced by this PR, just now more load-bearing for the anti-DoS story.

## Test coverage

Good — new/updated tests cover the ticket-issuance routes, the QR refresh-window rejection, and the personal-QR refresh interval. The camera-scanner behavioral change (issue #2 above) is the one meaningful gap.

src/client/api/upload.ts
```diff
@@ -1,28 +1,58 @@
...
   try {
-    return await httpClient.post<{ id: string; url: string }>(
-      "/storage/avatar",
-      form
-    );
+    const ticket = await httpClient.post<UploadTicket>(endpoint, {
+      contentType: file.type,
+      size: file.size,
+    });
+    const { error } = await createClient()
```
dinisjcorreia, line 32: [P2] Preserve the failed-upload contract

`uploadToStorage` callers use `null` to show an upload failure, but `uploadToSignedUrl` returns normal rejected/expired-ticket and Storage failures as `{ error }`. Re-throwing that as a generic `Error` bypasses the `HttpClientError` catch, so avatar dialogs can remain loading (`ProfileSection`/`SettingsSection`) and signup only logs the failure. Return `null` here (and log if needed) so this path keeps the previous failure behavior.

dinisjcorreia (changes requested):
Request changes: one P2 error-handling regression in direct Storage uploads; details inline.

david-fs-valente (approved):
Reviewed the current head (`c901afa`) directly: `pnpm test` (269/269), `pnpm typecheck`, and `pnpm lint` (0 errors) all pass locally, and the earlier Copilot/`david-fs-valente`/`dinisjcorreia` findings (mojibake string, blocking camera-error regression, swallowed Storage-upload failures) all look resolved in the final commits.

A few small, non-blocking things I noticed that weren't already covered by prior review rounds:

1. **`QRCodeScanner`'s camera error is now a one-time toast, not fully silent.** The `2026-08-06` commit fixed the blocking regression (`<video>` is always rendered now), but `onError` still calls `toast.error(...)` once even though the surrounding comment explicitly documents that some mobile browsers fire this error while the camera keeps working fine. So a user on one of those browsers now sees a "couldn't access the camera" toast on a scanner that's actually working — better than before (no more full breakage), but not back to the original fully-silent behavior for that known false-positive case.
2. `src/components/QRCode/QRCodeTab/ProfileTab/index.test.tsx` has a new Prettier import-order warning (`pnpm lint` — still 0 errors overall, just a new warning).
3. The 30s freshness threshold (`actionQrCodeRefreshRateMs * 2`) is computed independently in both `actionService.ts` (JWT `expiresIn`) and `actions/[id]/route.ts` (the new freshness check) — worth a shared constant so they can't drift apart later.
4. Test coverage gap: `src/client/api/upload.test.ts` only covers the Storage-rejection path for `uploadAvatar`; no test for the browser-side signature-mismatch rejection, the success path, or the `decoded.timestamp > Date.now()` branch in the actions route (only the "too old" and exact-`Date.now()` cases are tested).

None of these block merging — approving.

PR #277, opened 2026-07-29
ci: add local quality gates

### LINKED ISSUE #157: [CI] Add branch protection and PR workflow (main/dev, review, required CI)
### Description
Nothing enforces review or green CI before code reaches `main`, and `main` is directly pushable.

### Why
The CI gate has no teeth without protection.

### How
Adopt feat -> dev -> main. Phase A (now, no dependency): require PR + >=1 review + block direct push/force-push on `main` and `dev`. Phase B: require CI status checks. Add a `CODEOWNERS` file.

### Dependencies
Depends on #124 (CI gate) for the required-CI check (Phase A branch protection is independent).

hugo2006alm:
## What changed

- Added Husky + lint-staged: pre-commit formats and ESLints staged files; pre-push runs full lint, typecheck, and tests.
- Added repository-wide `CODEOWNERS` using current repository administrators.
- Stabilized Vitest on Windows with bounded workers and suitable test/hook timeouts.
- Made Prisma boundary scanning path-separator independent.
- Moved pnpm settings to `pnpm-workspace.yaml`, copied that file into Docker dependency builds, and updated dependency-alert remediations (`next`, `sharp`, Next ESLint packages, and SheetJS).

## Why

#157 requires review and CI controls. Repository already runs CI for pull requests; this adds fast commit-time checks, full push-time gates, and future code-owner policy.

Dependency updates are intentional security maintenance. Current SheetJS releases are distributed through SheetJS's exact-version CDN tarball rather than npm; this is pinned to `xlsx-0.20.3.tgz`, with accepted tradeoff that pnpm records no registry integrity hash for that external tarball.

## Validation

- `pnpm exec lint-staged`: configuration loaded; config/hook files do not match code-format tasks.
- `pnpm lint`: passed (existing Windows CRLF/Prettier warnings only).
- `pnpm typecheck`: passed.
- `pnpm test`: 63 files, 263 tests passed.
- CODEOWNERS: all 15 listed users verified as current repository admins.

## Remaining external setup

GitHub rejects branch protection/rulesets for this private repository with: "Upgrade to GitHub Pro or make this repository public to enable this feature." After that change, protect `main` and `dev`, require one review, require Code Owner review, block force pushes, and require CI checks. Coolify deployment is intentionally unchanged.


david-fs-valente:
**Non-blocking notes**

1. Pre-commit now runs the **full** typecheck (`next typegen && tsc --noEmit`) and **full** test suite (263 tests) on every commit — combined with the Windows hook-timeout bump to 90s for a single test, routine commits could get noticeably slow, especially on Windows. Worth considering scoping to affected files, or moving typecheck/test to a pre-push hook instead of pre-commit.
2. `lint-staged` only runs `prettier --write`, not `eslint` — CI's `pnpm lint` isn't mirrored locally, so a lint-only violation can still be committed and will only be caught later in CI.
3. `vitest.config.ts`'s new `hookTimeout: 90_000` now matches the explicit `}, 90_000)` override already present on the one test in `student/[...data]/page.test.tsx` — that per-test override is redundant now (equal to the new default) and could be dropped, though leaving it with its explanatory comment is a reasonable call too.
4. `CODEOWNERS` is a single `*` rule naming ~15 individuals rather than a scoped team — already called out in the file's own comment and the PR body as a placeholder pending an org team, so not a real issue, just flagging it's inert until branch protection can actually be enabled.

Verified good: the `boundaries.test.ts` path-separator fix is correct (old `/`-hardcoded check never matched on Windows, so repo files weren't excluded from the Prisma boundary scan there), and husky's `prepare` script is safe for the Docker build since `.dockerignore` excludes `.git` and husky v9 no-ops when there's no git repo present.

david-fs-valente:
## Review

**What it does:** Adds Husky pre-commit (`lint-staged` + `typecheck` + `test`), `lint-staged` (Prettier on staged files), a `CODEOWNERS` file, bumps `actions/checkout` to v5, migrates pnpm's `overrides`/`onlyBuiltDependencies` out of `package.json` into a new `pnpm-workspace.yaml` (with `Dockerfile` updated to `COPY` it), fixes the Prisma-boundary test to be path-separator-independent (Windows fix), and raises various Vitest timeouts/worker caps for Windows stability.

### Strengths
- The `boundaries.test.ts` fix (`file.includes("/application/repositories/")` → a `relative()` + `path.sep`-based check) is a real, correctly-targeted Windows bug fix, and it incidentally widens `.test.ts` exclusion to `.test.tsx` too (harmless, arguably more correct).
- Moving pnpm config into `pnpm-workspace.yaml` and updating the `Dockerfile`'s `COPY` line together is consistent — the overrides weren't dropped, just relocated to match newer pnpm's workspace-config convention, and the two new override entries (`brace-expansion`, `sharp`) are present in the new file.

1. **Scope creep, undocumented in the PR body.** The description only mentions Husky/lint-staged, CODEOWNERS, Vitest/Windows stability, and the boundary-test fix — but the diff also bumps `next` (15.5.20→21), `sharp` (0.34.5→0.35.3, plus new native-binary variants), `eslint-config-next`/`@next/eslint-plugin-next`, and swaps `xlsx` from the npm registry (`^0.18.5`) to a **direct CDN tarball URL** (`https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz`). None of this is explained — worth confirming whether it was intentional or should be split out.
2. **The `xlsx` tarball has no integrity hash in the lockfile** (`resolution: {tarball: <url>}` with no `integrity:` field) — pnpm can't verify the bytes downloaded from `cdn.sheetjs.com` haven't changed. This is SheetJS's own documented install method (they don't publish current versions to npm), so it may be an accepted tradeoff, but it's worth flagging as a supply-chain consideration for a repo that otherwise pins everything through the npm registry with integrity hashes.
3. `CODEOWNERS` lists 15 hardcoded GitHub usernames as owners of `*`. Worth a sanity check that all of them are still active collaborators with write access — a stale/mistyped username would silently break the "require Code Owner review" branch rule this is meant to enable.
4. Pre-commit now runs the full `pnpm test` + `pnpm typecheck` on every commit. Consistent with "equivalent local quality gates" as CI, but will make every local commit noticeably slower (263 tests) — just flagging the tradeoff.

### Test coverage
N/A (tooling-only) aside from the boundary-test fix itself, which is well-targeted.

david-fs-valente (commented):
## Overview

Adds Husky + lint-staged (pre-commit format/lint, pre-push full lint+typecheck+test), a CODEOWNERS file, bumps `actions/checkout` to v5, moves pnpm config into `pnpm-workspace.yaml`, and does dependency-alert remediation (`next`, `sharp`, ESLint-Next packages, SheetJS `xlsx`).

## Issues / risks

1. **The git hooks are not executable — they will silently never run.** Checked the actual tree: both `.husky/pre-commit` and `.husky/pre-push` have git mode `100644` (regular file), not `100755`. Husky v9 sets `core.hooksPath` directly at the hook files themselves (no `_` wrapper anymore), so git executes them directly — a non-executable hook file is silently skipped by git on any POSIX system. As committed, **neither the pre-commit format/lint step nor the pre-push lint/typecheck/test gate actually fires**, defeating the PR's purpose. This also explains why the PR's own validation note ("`pnpm exec lint-staged`: configuration loaded...") ran the command manually rather than via an actual `git commit` — that end-to-end path was never exercised. Fix: `chmod +x .husky/pre-commit .husky/pre-push` and re-commit.
2. **`xlsx` dependency has no integrity hash.** Moving off npm to `https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz` is a reasonable response to SheetJS's own distribution changes, but the lockfile entry (`resolution: {tarball: ...}`) has no `integrity:` field at all, unlike every other package. A future `pnpm install` fetches that URL with zero verification that the bytes match what was reviewed here — if the CDN is compromised or serves different content, nothing would catch it. pnpm/npm support integrity-pinned tarball URLs via a `#sha512-...` fragment; consider adding that (SheetJS publishes hashes) rather than accepting an unpinned external URL as a permanent tradeoff.
3. Minor/coordination note: this PR's changes to `page.test.tsx`, `boundaries.test.ts`, and `vitest.config.ts` (the 90s import timeout, path-separator-independent boundary check, `maxWorkers`/timeouts) are **byte-for-byte identical** to changes already in #276. Since both branch off `dev` independently, this is duplicated effort from the same author across two PRs — should merge cleanly (git resolves identical target content without conflict), but worth landing one and rebasing the other rather than shipping the same diff twice.
4. Minor: `sharp` is bumped to `^0.35.3`, which per the lockfile requires Node `>=20.9.0`, but this PR doesn't add an `engines` field (that only shows up in #278, which isn't stacked on #277). Docker/CI already pin Node 20, so risk is low, but a contributor on an older local Node 20.x could hit a broken `sharp` install with no clear signal until #278 also lands.

## Strengths

CODEOWNERS is a sensible placeholder with a clear comment on how to replace it later; CI checkout bump is trivial and safe; moving pnpm config to `pnpm-workspace.yaml` (and copying it into the Docker deps stage) is correctly wired so Docker installs still see the overrides.

dinisjcorreia (approved):
Reviewed current head f50d86e. No blocking findings.

Verified: fresh frozen install; lint (existing warnings only); typecheck; 263 tests passed. Husky hooks are executable and installed.

PR #278, opened 2026-07-29
test: add event readiness harness

hugo2006alm:
## What changed

- Add Playwright with local health smoke coverage across desktop Chrome and mobile browser projects.
- Add opt-in staging student QR and direct CV signed-upload checks.
- Add guarded k6 event-peak harness for health, QR issuance, and upload-ticket load; reject settings that exceed per-student ticket limits.
- Document setup, staging-only credentials, generated artifacts, and storage-state credential handling.
- Keep Vitest isolated from Playwright specs and stable on Windows.

## Why

QR and upload flows need repeatable browser checks and controlled staging load before event. Authenticated Playwright flows and k6 harness refuse to run unless non-production use is explicitly confirmed.

## Validation

- `pnpm lint --quiet`: passed after merging current #276 head.
- `pnpm typecheck`: passed after merging current #276 head.
- `pnpm test`: 64 files, 267 tests passed after merging current #276 head.
- `pnpm test:e2e --project=chromium`: 1 health smoke passed; 2 staging-only checks skipped.
- k6 binary is not installed locally; script execution is documented for staging.

## Merge order

This PR is stacked on #276 and should merge after it, then be retargeted to `dev`.


david-fs-valente:
**Non-blocking notes**

1. **Validation gap** — the PR's "Validation" section lists `test:e2e` and `typecheck` but not `pnpm test`. This PR specifically had to patch `vitest.config.ts` to exclude `tests/e2e/playwright/**` from Vitest's auto-discovery (Vitest's default `include` glob matches `*.spec.ts` too, so without the exclusion `pnpm test`/CI would try to run the new Playwright specs under Vitest's runner and fail). That's exactly the kind of change that needs a recorded `pnpm test` run before merge to confirm no interference with the regular suite.
2. **Asymmetric production guardrail** — the k6 harness refuses to run without `CONFIRM_NON_PRODUCTION=yes`, but `student-event-flow.spec.ts` (which also performs real, state-changing calls — minting a CV upload ticket and writing an actual file to Storage) has no equivalent runtime guard, only a doc warning ("Never use a production account or a production URL"). A mistyped `E2E_BASE_URL` plus `E2E_ALLOW_UPLOAD_TICKETS=yes` could hit production with nothing in code stopping it. Worth mirroring the k6 script's explicit-confirmation pattern here too.
3. Minor: `tests/e2e/.staging-*.json` (the captured Playwright storage-state file) is correctly gitignored, but the README doesn't flag that it holds a live staging session and should be handled like a credential.

Also worth confirming as a process note (not a defect): this PR is stacked on #276's branch rather than `dev`, so it can't land until #276 does and should be retargeted at `dev` once #276 merges.

Verified good: the `student-event-flow.spec.ts` CV-upload test drives the real client path (`uploadToSignedUrl` against a `/api/storage/cv` ticket), giving genuine end-to-end coverage of the new signed-upload flow from #276.

david-fs-valente:
## Review

Note: this PR's base branch is `chore/qr-upload-resilience-audit` (#276's branch), not `dev` — it's stacked on #276 and should be merged after (or together with) it.

**What it does:** Adds Playwright (always-on health smoke test; staging-gated authenticated QR + CV-upload-ticket tests) and a k6 load-test script for pre-event capacity checks, plus docs.

### Strengths
- Good safety posture throughout: both the Playwright staging tests and the k6 script hard-refuse to run without `CONFIRM_NON_PRODUCTION=yes`, and the k6 script throws immediately if `E2E_BASE_URL`/`ACTION_ID`/`STUDENT_COOKIES` are missing for the scenario selected.
- Verified the endpoints/response shapes the new tests assert against match reality: `/api/health` → `{status: 200}`, `/api/qrcode` → `{data: string}`, `GET /api/actions/[id]` → includes `qrCode`.
- Correctly adds `tests/e2e/playwright/**` to Vitest's `exclude` — Vitest's default include pattern matches `*.spec.ts` too, so without this the new Playwright specs would otherwise get picked up and fail under Vitest's runner. Real fix, not a redundant line.
- `.gitignore` additions (`.staging-*.json`, Playwright artifact dirs) are correctly scoped, and the README is explicit the staging storage-state file should be handled like a credential.

1. The new `test:e2e`/`test:load` scripts aren't wired into CI (`.github/workflows/ci.yml` untouched) — consistent with the PR's stated scope ("local health smoke coverage"), but worth a follow-up if the intent is eventually to gate PRs on at least the health check.
2. The k6 `upload-tickets` scenario relies on the operator manually keeping `STUDENT_COOKIES` pool size / VU count / duration under each account's 5-tickets/minute budget — documented in the README, not enforced in code, so it's easy to misconfigure into a wall of 429s that just tests the rate limiter rather than real throughput. Not a bug, just a footgun worth a comment in the script itself.

### Test coverage
This *is* the test coverage addition — appropriately scoped, well-guarded against accidental prod usage.

david-fs-valente (commented):
## Overview

Adds a Playwright suite (health smoke + opt-in authenticated staging flows for QR and CV upload) and a guarded k6 load-testing harness for event-day traffic, with explicit non-production confirmation gates throughout.

Note on merge order: this PR's base is `chore/qr-upload-resilience-audit` (#276's branch), not `dev`, as the description states — it'll need to be retargeted after #276 merges. Its diff is small and clean specifically because it doesn't duplicate #276's changes.

## Strengths

- Good safety posture: both the authenticated Playwright spec and the k6 script hard-refuse to run without `CONFIRM_NON_PRODUCTION=yes`, and the k6 script proactively rejects load configurations that would exceed the app's real 5-tickets/minute/student rate limit (`ticketsPerStudentPerMinute` calculation) rather than just letting the load test generate a wall of false-positive 429s.
- `.gitignore` correctly excludes generated Playwright artifacts and the staging storage-state file (documented as credential-like, never to be committed/shared).
- `vitest.config.ts` excludes `tests/e2e/playwright/**` so Vitest doesn't try to execute Playwright specs — correct isolation.
- The k6 QR-scenario check (`res.json("qrCode")`) matches the real route's response field name (`getActionQrCode` returns `{ action, qrCode }`).
- CV upload test correctly uses a bare Supabase anon client for `uploadToSignedUrl` (no auth session needed — the signed token itself is the credential), consistent with #276's new upload architecture.

## Minor suggestions (not blocking)

- `tests/load/event-readiness.js`'s `default()` function has no `else` branch — an unrecognized `K6_SCENARIO` value silently runs zero requests and "passes" (thresholds are vacuously satisfied). A guard at startup (`if (!["health","qr","upload-tickets"].includes(scenario)) throw ...`) would fail loudly instead.
- Passing `STUDENT_COOKIES` via env var is fine for a staging-only, explicitly-gated tool, but worth a one-line reminder in the README that these are live session cookies and shouldn't linger in shell history — the doc already treats the storage-state file this way, could extend the same note to this env var.
- As flagged in #276's review: this PR is a natural place to also assert that the Storage bucket actually rejects an oversized/mismatched-type direct upload (the one real enforcement boundary in the new architecture), but it currently only exercises the happy path (valid small PDF succeeds).

tests/e2e/playwright/student-event-flow.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    "Set E2E_STUDENT_STORAGE_STATE to a staging student Playwright storage-state file."
+  );
+  test.skip(
+    !confirmNonProduction,
+    "Set CONFIRM_NON_PRODUCTION=yes; never run authenticated event tests against production."
+  );
+
+  test.use({ storageState: storageState! });
+
+  test("student QR code is generated", async ({ request }) => {
```
dinisjcorreia: [P1] Built-in `request` fixture is isolated `APIRequestContext`; `test.use({ storageState })` initializes browser contexts only. These requests omit saved Supabase session cookie, so `/api/qrcode` and upload-ticket requests return 401 when opt-in suite runs. Use `page.request`/browser-context request, or create API request context explicitly with storage state.

tests/e2e/playwright/student-event-flow.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const response = await request.get("/api/qrcode");
+
+    expect(response.status()).toBe(200);
+    await expect(response.json()).resolves.toEqual({
+      data: expect.any(String),
+    });
+  });
+
+  test("student can obtain and use a CV upload ticket", async ({ request }) => {
```
dinisjcorreia: [P1] After auth fix, both upload flags make suite issue nine tickets with one student: three projects × (this test + two restriction tickets). Route caps `session.id` at five tickets/minute; projects run parallel, so at least four requests get 429 and documented staging run fails. Run ticket coverage in one project, use separate accounts, or keep aggregate under five/minute.

dinisjcorreia (changes requested):
Two blocking issues in opt-in authenticated staging flow.

david-fs-valente (approved):
Reviewed — looks solid. Cross-checked the Playwright/k6 assertions against the actual route implementations (health, qrcode, actions/[id], storage/cv) and the rate-limit/size config; everything lines up, and the staging-only gating (CONFIRM_NON_PRODUCTION, opt-in flags, gitignored storage-state) is appropriately cautious.

One note: `.github/workflows/ci.yml` isn't updated, so `pnpm test:e2e` (the Playwright health smoke test) doesn't run in CI yet — it's local/manual only for now. Worth a follow-up if we want that enforced automatically before the event.

PR #296, opened 2026-08-07
fix: orphaned Supabase Auth accounts on deletion

### LINKED ISSUE #285: [Fix] Close orphaned-account gap between Supabase Auth and public.User on delete
### Description
There's no cascade between Supabase Auth's `auth.users` and `public.User` — deletes orphan in both directions. Within `public`, Prisma's own cascades are fine (`Student`/`Employee` → `User`, `Employee` → `Company`, `SavedStudent` → `Student`/`Employee`/`Company`); the gap is specifically between `auth.users` and `public.User`, correlated purely by convention (`public.User.id` == `auth.users.id` at signup, never DB-enforced).

### Why
This is a two-way gap, and one direction is already live, not hypothetical. **App → Auth**: every delete path in the app (`deleteStudentForAdmin`, `deleteEmployeeForAdmin`, `deleteAdminAccount`) calls the repository's plain `prisma.user.delete()` and never calls Supabase Auth's `admin.auth.admin.deleteUser()` — the only place that's ever called today is signup-failure rollback (`authApplicationService.ts`). So clicking "delete" in the app's own admin backoffice already leaves an orphaned `auth.users` row every time. **Auth → App**: deleting a user from the Supabase Auth dashboard leaves `public.User`/`Student`/`Employee` fully intact — a permanently un-loginable "real" account forever, and if the same email signs up again, Supabase issues a new `auth.users.id` that won't match the old orphaned row. `auth.users` belongs to Supabase's own schema, which Prisma doesn't model, so neither direction can be expressed as an ordinary `onDelete: Cascade`.

### How
1. Close the App → Auth direction (the one actively firing today): add the missing `admin.auth.admin.deleteUser()` call to `deleteStudentForAdmin`, `deleteEmployeeForAdmin`, and `deleteAdminAccount` — centralize it once in the shared `deleteUser()` repository function or a wrapping service function all three call, mirroring the existing signup-rollback pattern.
2. Decide on the Auth → App direction: a Postgres trigger on `auth.users` delete that removes the matching `public.User` row (closes it for good, but touches Supabase-owned schema), or documenting "always delete through the app, never the Supabase dashboard directly" as an operational rule.
3. Audit for existing orphaned rows on both sides in current environments before/while fixing this, since the gap has presumably already produced some.

### Dependencies
None.

dinisjcorreia:
## Summary

- delete the Supabase Auth identity before deleting the matching Prisma user
- route student, employee, admin, and unconfirmed-account cleanup through the shared service
- document backoffice-only deletion as the Auth-to-App operational rule
- add a read-only SQL audit for orphaned rows in both directions

## Verification

- `pnpm test` (267 passed)
- `pnpm typecheck`
- `pnpm lint` (0 errors; 16 existing warnings)
- started `pnpm dev`; all three admin DELETE routes returned `401 {"error":"Unauthorized"}` without a session

Successful deletion was not exercised against a live account because this worktree has no configured Supabase environment or admin session.

## Manual deployment step

Run `supabase/audit-orphaned-accounts.sql` manually in every Supabase environment before deployment. Investigate each reported row before deleting anything; the query is read-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ backoffice, which removes the Supabase Auth identity before the application
 row. Never delete users directly in the Supabase Auth dashboard: that leaves an
 unloginable application account.
 
-Before deploying this fix, and after any manual Auth operation, run
+Before deploying changes that affect account deletion, and after any manual
+Auth operation, run
 [`supabase/audit-orphaned-accounts.sql`](./supabase/audit-orphaned-accounts.sql)
 manually in every environment's Supabase SQL editor. It is read-only and reports
 orphans in both directions; investigate each result before deleting anything.
--- a/README.md
+++ b/README.md
@@ -29,6 +29,19 @@ through Supabase Auth. Application tables must not store passwords or password
 reset tokens, and application routes must not provide separate password-change
 flows.
 
+#### Account deletion
+
+`auth.users.id` and `public."User".id` match by convention; no foreign key can
+span Supabase Auth and Prisma. Always delete accounts through the admin
+backoffice, which removes the Supabase Auth identity before the application
+row. Never delete users directly in the Supabase Auth dashboard: that leaves an
+unloginable application account.
+
+Before deploying this fix, and after any manual Auth operation, run
+[`supabase/audit-orphaned-accounts.sql`](./supabase/audit-orphaned-accounts.sql)
+manually in every environment's Supabase SQL editor. It is read-only and reports
+orphans in both directions; investigate each result before deleting anything.
+
 ---
 
 # Getting Started
--- a/src/application/services/authApplicationService.ts
+++ b/src/application/services/authApplicationService.ts
@@ -90,6 +90,14 @@ export async function rollbackAuthUser(userId: string) {
   }
 }
 
+// Thrown only when the Supabase Auth deletion call itself failed - i.e.
+// neither the Auth identity nor the app row has been touched yet. Callers
+// that catch a deleteUserAccount() failure must check for this specifically:
+// it's the one case where falling back to an app-only delete would create an
+// Auth-only orphan. Extends HttpError so existing `instanceof HttpError`
+// error-response mapping in every other caller keeps working unchanged.
+export class AuthAccountDeletionError extends HttpError {}
+
 // Supabase Auth and public.User cannot share a foreign key. Delete Auth first
 // so an Auth API failure cannot remove the app row and leave a login identity
 // that the backoffice can no longer retry deleting.
@@ -98,7 +106,7 @@ export async function deleteUserAccount(userId: string) {
   // Only an explicit code proves absence. Missing codes can also mean the
   // malformed self-hosted GoTrue failures described above, so stay fail-closed.
   if (error && error.code !== "user_not_found")
-    throw new HttpError(
+    throw new AuthAccountDeletionError(
       usableAuthErrorMessage(error?.message, "Unable to delete account"),
       500
     );
@@ -200,8 +208,25 @@ export async function completeOAuthSignIn(input: {
     try {
       await deleteUserAccount(existingByEmail.id);
     } catch (error) {
-      // ponytail: verified login wins here; audit catches any Auth-only
-      // orphan. Add a retry queue if these failures become frequent.
+      if (error instanceof AuthAccountDeletionError) {
+        // The Auth deletion call itself failed - neither the Auth identity
+        // nor the app row has been touched, so deleting only the app row
+        // here would create an Auth-only orphan. Leave both sides intact
+        // and let this sign-in attempt fail; the OAuth callback route
+        // catches this and redirects to /auth/auth-code-error, so the user
+        // just retries once Auth is reachable again.
+        reportError(
+          error,
+          { operation: "delete_unconfirmed_account_placeholder" },
+          "Failed to delete unconfirmed account placeholder's Auth identity"
+        );
+        throw error;
+      }
+
+      // Auth was already confirmed deleted or absent (deleteUserAccount only
+      // reaches this point after that) - e.g. a raced concurrent OAuth
+      // callback deleted the app row first. No orphan risk: finish the
+      // cleanup and continue provisioning.
       reportError(
         error,
         { operation: "delete_unconfirmed_account_placeholder" },
--- a/src/application/services/authApplicationService.ts
+++ b/src/application/services/authApplicationService.ts
@@ -94,7 +94,11 @@ export async function rollbackAuthUser(userId: string) {
 // that the backoffice can no longer retry deleting.
 export async function deleteUserAccount(userId: string) {
   const { error } = await createAdminClient().auth.admin.deleteUser(userId);
-  if (error) throw new HttpError(error.message, 500);
+  if (error && error.code !== "user_not_found")
+    throw new HttpError(
+      usableAuthErrorMessage(error?.message, "Unable to delete account"),
+      500
+    );
   await deleteUser(userId);
 }
 
@@ -190,7 +194,18 @@ export async function completeOAuthSignIn(input: {
     // Unconfirmed - discard the dangling placeholder instead of relinking
     // onto it, so AuthNEI (institutionally-verified) can claim the email
     // fresh below rather than inheriting potentially attacker-planted data.
-    await deleteUserAccount(existingByEmail.id);
+    try {
+      await deleteUserAccount(existingByEmail.id);
+    } catch (error) {
+      // ponytail: verified login wins here; audit catches any Auth-only
+      // orphan. Add a retry queue if these failures become frequent.
+      reportError(
+        error,
+        { operation: "delete_unconfirmed_auth_placeholder" },
+        "Failed to delete unconfirmed Supabase Auth placeholder"
+      );
+      await deleteUser(existingByEmail.id);
+    }
   }
 
   // First AuthNEI sign-in for this identity, no (trustworthy) pre-existing
--- a/src/application/services/authApplicationService.ts
+++ b/src/application/services/authApplicationService.ts
@@ -13,6 +13,7 @@ import {
 import { withTransaction } from "../repositories/transaction";
 import {
   deleteUser,
+  deleteUserIfExists,
   findUserByEmail,
   findUserSessionByEmail,
   findUserSessionById,
@@ -94,6 +95,8 @@ export async function rollbackAuthUser(userId: string) {
 // that the backoffice can no longer retry deleting.
 export async function deleteUserAccount(userId: string) {
   const { error } = await createAdminClient().auth.admin.deleteUser(userId);
+  // Only an explicit code proves absence. Missing codes can also mean the
+  // malformed self-hosted GoTrue failures described above, so stay fail-closed.
   if (error && error.code !== "user_not_found")
     throw new HttpError(
       usableAuthErrorMessage(error?.message, "Unable to delete account"),
@@ -201,10 +204,10 @@ export async function completeOAuthSignIn(input: {
       // orphan. Add a retry queue if these failures become frequent.
       reportError(
         error,
-        { operation: "delete_unconfirmed_auth_placeholder" },
-        "Failed to delete unconfirmed Supabase Auth placeholder"
+        { operation: "delete_unconfirmed_account_placeholder" },
+        "Failed to fully delete unconfirmed account placeholder"
       );
-      await deleteUser(existingByEmail.id);
+      await deleteUserIfExists(existingByEmail.id);
     }
   }
 
--- a/src/application/services/authApplicationService.ts
+++ b/src/application/services/authApplicationService.ts
@@ -89,6 +89,15 @@ export async function rollbackAuthUser(userId: string) {
   }
 }
 
+// Supabase Auth and public.User cannot share a foreign key. Delete Auth first
+// so an Auth API failure cannot remove the app row and leave a login identity
+// that the backoffice can no longer retry deleting.
+export async function deleteUserAccount(userId: string) {
+  const { error } = await createAdminClient().auth.admin.deleteUser(userId);
+  if (error) throw new HttpError(error.message, 500);
+  await deleteUser(userId);
+}
+
 // Defense-in-depth alongside User.active, which is what getServerSession()
 // actually enforces on every request - this additionally stops the person
 // from obtaining a *new* Supabase session at all once deactivated. Best-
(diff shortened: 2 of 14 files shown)


src/application/services/authApplicationService.ts
```diff
@@ -89,6 +89,15 @@ export async function rollbackAuthUser(userId: string) {
   }
 }
 
+// Supabase Auth and public.User cannot share a foreign key. Delete Auth first
+// so an Auth API failure cannot remove the app row and leave a login identity
+// that the backoffice can no longer retry deleting.
+export async function deleteUserAccount(userId: string) {
+  const { error } = await createAdminClient().auth.admin.deleteUser(userId);
+  if (error) throw new HttpError(error.message, 500);
```
david-fs-valente: [P2] Pre-existing orphaned app-only rows become permanently undeletable

`deleteUserAccount()` fails closed: if `admin.auth.admin.deleteUser()` errors, the app row is never deleted either. `supabase/audit-orphaned-accounts.sql`'s second query lists exactly the rows where this will always happen — app rows with no matching Auth user, i.e. orphans that predate this fix. An admin who finds one via the audit and clicks delete in the backoffice gets a 500 every time, with no path left to remove it — even though the README now says deletion must go through the backoffice only.

Suggested fix: when the Auth user is already missing, skip straight to deleting the app row instead of throwing (or add a backoffice affordance to delete just the app-side row for rows already known to be Auth-orphaned). Worth a test covering "Auth user not found" as its own case, distinct from "Auth API errored."

src/application/services/authApplicationService.ts
```diff
@@ -181,7 +190,7 @@ export async function completeOAuthSignIn(input: {
     // Unconfirmed - discard the dangling placeholder instead of relinking
     // onto it, so AuthNEI (institutionally-verified) can claim the email
     // fresh below rather than inheriting potentially attacker-planted data.
-    await deleteUser(existingByEmail.id);
+    await deleteUserAccount(existingByEmail.id);
```
david-fs-valente: [P1] Cleanup of a spoofed placeholder can now block a legitimate student's login

This discards an attacker-plantable placeholder (someone hitting Supabase's signup API directly under a real student's ISEP email before that student ever logs in — see the comment above `isEmailConfirmed`). Since this now goes through `deleteUserAccount()`, a transient failure deleting *that* Auth identity throws, and `app/auth/callback/route.ts` catches it and redirects the real student to the generic auth-error page — coupling cleanup of someone else's fake row to their own login. Before this PR, this leg only touched the app table and couldn't realistically fail.

Suggested fix: make this cleanup best-effort here, matching `rollbackAuthUser`/`setAuthUserBanned` elsewhere in this file — catch, `reportError`, and continue provisioning the real account rather than throwing. Worth a test asserting login still succeeds when this delete fails.

david-fs-valente (changes requested):
Request changes: two issues in the Auth-first deletion flow — an orphaned-row cleanup gap and a login availability regression; details inline.

src/application/services/authApplicationService.ts
```diff
@@ -89,6 +89,15 @@ export async function rollbackAuthUser(userId: string) {
...
 }
 
+// Supabase Auth and public.User cannot share a foreign key. Delete Auth first
+// so an Auth API failure cannot remove the app row and leave a login identity
+// that the backoffice can no longer retry deleting.
+export async function deleteUserAccount(userId: string) {
+  const { error } = await createAdminClient().auth.admin.deleteUser(userId);
+  if (error) throw new HttpError(error.message, 500);
+  await deleteUser(userId);
+}
```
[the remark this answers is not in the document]
dinisjcorreia: Fixed in dbcea30. deleteUserAccount now passes Supabase errors through usableAuthErrorMessage with a stable "Unable to delete account" fallback. Added an opaque-error regression test.

README.md
```diff
@@ -29,6 +29,19 @@ through Supabase Auth. Application tables must not store passwords or password
...
+`auth.users.id` and `public."User".id` match by convention; no foreign key can
+span Supabase Auth and Prisma. Always delete accounts through the admin
+backoffice, which removes the Supabase Auth identity before the application
+row. Never delete users directly in the Supabase Auth dashboard: that leaves an
+unloginable application account.
+
+Before deploying this fix, and after any manual Auth operation, run
+[`supabase/audit-orphaned-accounts.sql`](./supabase/audit-orphaned-accounts.sql)
+manually in every environment's Supabase SQL editor. It is read-only and reports
+orphans in both directions; investigate each result before deleting anything.
```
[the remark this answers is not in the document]
dinisjcorreia: Fixed in dbcea30. README now uses durable wording: before deploying changes that affect account deletion.

src/application/services/authApplicationService.ts
dinisjcorreia: Fixed in dbcea30. Explicit Supabase user_not_found is now treated as an idempotent Auth deletion, so the app-only orphan is removed. Other Auth failures remain fail-closed. Added a dedicated regression test.

src/application/services/authApplicationService.ts
dinisjcorreia: Fixed in dbcea30. Placeholder Auth cleanup is now best-effort: failure is reported, then the app placeholder is deleted before provisioning the verified identity, avoiding the email uniqueness failure that swallowing the error alone would cause. Added a regression test proving login provisioning continues.

src/application/services/authApplicationService.ts
```diff
@@ -89,6 +89,19 @@ export async function rollbackAuthUser(userId: string) {
...
+// so an Auth API failure cannot remove the app row and leave a login identity
+// that the backoffice can no longer retry deleting.
+export async function deleteUserAccount(userId: string) {
+  const { error } = await createAdminClient().auth.admin.deleteUser(userId);
+  if (error && error.code !== "user_not_found")
+    throw new HttpError(
+      usableAuthErrorMessage(error?.message, "Unable to delete account"),
+      500
+    );
+  await deleteUser(userId);
```
david-fs-valente, line 113: **Confirmed regression.** `ActionCompletion.studentId` is `ON DELETE RESTRICT` (`prisma/migrations/20260712000000_init/migration.sql:152`), while `Student` cascades from `User` via `ON DELETE CASCADE`. Since `deleteUserAccount()` now deletes the Supabase Auth identity *before* reaching this line, deleting any student who has completed at least one QR action will succeed at removing their Supabase Auth identity, then fail right here with a Postgres FK violation when the cascade from `User` hits the restricted `ActionCompletion` row.

Net effect: Auth identity permanently gone, `Student`/`User` row stuck in Postgres — exactly the "unloginable application account" orphan this PR's own README section warns about, now reachable through the normal admin-delete flow for any student who scanned an action (a common case, not an edge case).

Before this PR, `deleteStudentForAdmin` called `deleteUser` directly with no Auth step, so this same FK conflict was a harmless no-op on both sides. The reordering turns it into a destructive partial failure.

Suggest clearing/handling `ActionCompletion` rows for the student inside a transaction before this delete (or reconsidering the FK's `ON DELETE` behavior) so this path can't leave an Auth-deleted, DB-row-stuck orphan.

src/application/services/authApplicationService.ts
```diff
@@ -181,7 +194,18 @@ export async function completeOAuthSignIn(input: {
...
+      await deleteUserAccount(existingByEmail.id);
+    } catch (error) {
+      // ponytail: verified login wins here; audit catches any Auth-only
+      // orphan. Add a retry queue if these failures become frequent.
+      reportError(
+        error,
+        { operation: "delete_unconfirmed_auth_placeholder" },
+        "Failed to delete unconfirmed Supabase Auth placeholder"
+      );
+      await deleteUser(existingByEmail.id);
```
david-fs-valente: This fallback `await deleteUser(existingByEmail.id);` isn't wrapped in a try/catch. If it throws — e.g. a raced/duplicate OAuth callback where a concurrent call already deleted this row (Prisma P2025) — the error escapes uncaught here, aborting what should otherwise be a successful AuthNEI sign-in, rather than being absorbed the way the surrounding comment ("verified login wins here") implies.

src/application/services/authApplicationService.ts
```diff
@@ -89,6 +89,19 @@ export async function rollbackAuthUser(userId: string) {
   }
 }
 
+// Supabase Auth and public.User cannot share a foreign key. Delete Auth first
+// so an Auth API failure cannot remove the app row and leave a login identity
+// that the backoffice can no longer retry deleting.
+export async function deleteUserAccount(userId: string) {
+  const { error } = await createAdminClient().auth.admin.deleteUser(userId);
+  if (error && error.code !== "user_not_found")
```
david-fs-valente, line 108: This `user_not_found` bypass depends on `error.code` being populated by supabase-js from the response body. `usableAuthErrorMessage`'s own comment above documents that self-hosted GoTrue can return a malformed/empty error body on infra-level failures — in that case supabase-js's `handleError()` leaves `errorCode` `undefined` (verified against the installed `@supabase/auth-js` source), so this check falls through to the generic throw instead of treating it as an already-deleted-elsewhere case. Narrow edge case, but worth a comment at least, since it's the same GoTrue quirk already called out nearby.

src/application/services/authApplicationService.ts
```diff
@@ -181,7 +194,18 @@ export async function completeOAuthSignIn(input: {
     // Unconfirmed - discard the dangling placeholder instead of relinking
     // onto it, so AuthNEI (institutionally-verified) can claim the email
     // fresh below rather than inheriting potentially attacker-planted data.
-    await deleteUser(existingByEmail.id);
+    try {
+      await deleteUserAccount(existingByEmail.id);
+    } catch (error) {
+      // ponytail: verified login wins here; audit catches any Auth-only
+      // orphan. Add a retry queue if these failures become frequent.
+      reportError(
```
david-fs-valente, line 230: This `reportError` always attributes the failure to "Failed to delete unconfirmed Supabase Auth placeholder," but `deleteUserAccount()` can also throw from the later `deleteUser(userId)` Prisma call after the Auth deletion already succeeded (see the FK comment above). In that case the log points investigators at the wrong side of the mismatch — worth branching the message (or noting which step failed) so on-call doesn't waste time checking the Auth side when the real leftover is the DB row.

supabase/audit-orphaned-accounts.sql
```diff
@@ -0,0 +1,16 @@
+-- Run manually in each environment's Supabase SQL editor. Read-only: this
+-- reports account mismatches but intentionally does not delete either side.
+
+-- Supabase Auth identities with no application account.
+select auth_user.id, auth_user.email
```
david-fs-valente: Minor operational gotcha: Supabase's SQL Editor generally only surfaces the result of the *last* statement when multiple `SELECT`s are pasted and run together. If an operator follows the README instructions and runs this whole file at once, this first query's result (Auth identities with no application account) may not be visible, only the second one. Consider noting in the header comment that each `SELECT` should be run individually, or merge both into a single query (e.g. `FULL OUTER JOIN` with a `direction` label column) so one run surfaces both directions.

david-fs-valente (changes requested):
## Review summary

Reviewed this directly: checked out `fix/close-orphaned-account-gap`, ran `pnpm lint`, `pnpm typecheck`, and `pnpm test` locally rather than trusting the PR description's numbers alone — all confirmed: 0 lint errors / 16 pre-existing warnings (none in touched files), typecheck clean, 270/270 tests passing.

The core idea — delete the Supabase Auth identity before the Prisma row, route all admin deletion paths through one shared `deleteUserAccount()`, document the operational rule, and add a read-only audit query — is sound, and the paths it does test are well covered.

However, one finding below is a concrete regression that undermines the PR's own stated goal, confirmed against the actual FK constraints in `prisma/migrations/20260712000000_init/migration.sql`. Requesting changes on that basis. The remaining comments are lower-severity but worth addressing before merge.

hugo2006alm (changes requested):
[P1] `deleteUserAccount` deletes Auth before Prisma, but `ActionCompletion.studentId` is `ON DELETE RESTRICT`. Deleting any student with a completion removes their Auth identity then fails deleting `User`, leaving an unloginable orphan. Add a cascade migration (or delete completions atomically before Auth deletion) before merge.

[P2] Placeholder cleanup can still throw when the fallback `deleteUser` races with another callback; catch/absorb Prisma P2025 before provisioning the verified identity.

src/application/services/authApplicationService.ts
dinisjcorreia: Fixed in da66e8a. ActionCompletion.student now uses onDelete: Cascade, with a committed Prisma migration replacing the restrictive FK and a regression test checking both schema and migration SQL. Migration was not executed locally.

src/application/services/authApplicationService.ts
dinisjcorreia: Fixed in da66e8a. Placeholder fallback now calls repository deleteUserIfExists(), implemented with deleteMany(), so a concurrent deletion is idempotent and cannot raise P2025.

src/application/services/authApplicationService.ts
dinisjcorreia: Addressed in da66e8a. Added an explicit comment documenting that only user_not_found proves absence; missing/malformed codes remain fail-closed because they may represent the self-hosted GoTrue infrastructure failure.

src/application/services/authApplicationService.ts
dinisjcorreia: Fixed in da66e8a. Logging now uses account-level operation/message wording (failed to fully delete unconfirmed account placeholder), covering failures from either Auth or Prisma without misdirecting investigation.

supabase/audit-orphaned-accounts.sql
dinisjcorreia: Fixed in da66e8a. Audit now uses one FULL OUTER JOIN query with an orphan_type label, so both mismatch directions appear in one SQL Editor result set.

dinisjcorreia:
Addressed latest requested changes in `da66e8a`:

- changed `ActionCompletion.studentId` to `ON DELETE CASCADE` via Prisma schema + migration, preventing Auth-first partial deletion for students with completions
- made raced OAuth placeholder cleanup idempotent with `deleteMany()`
- clarified fail-closed handling for malformed GoTrue errors and generalized mismatch logging
- merged orphan audit into one labeled result set

Verification: Prisma schema valid; 271 tests pass; typecheck clean; lint has 0 errors (16 pre-existing warnings); dev route check completed. Migration SQL is committed but was not executed locally.

hugo2006alm (changes requested):
[P1] Unconfirmed-account cleanup still creates an Auth-only orphan when coordinated deletion fails.

In completeOAuthSignIn, a deleteUserAccount failure is caught and followed by deleteUserIfExists(existingByEmail.id). That removes only public.User, continues provisioning, and knowingly leaves the old auth.users identity behind. This contradicts this PR's central guarantee that deletion paths go through the shared Auth-plus-app service; audit detection does not make the cleanup correct.

Please keep the app row until both sides can be deleted, or persist/retry a coordinated cleanup without deleting only the application side. The verified login availability requirement may need a durable cleanup/tombstone design rather than the current best-effort one-sided delete.

hugo2006alm:
Superseded by consolidated stabilization PR #305. All four commits were preserved as logical commit d4fdef9 with Dinis as author and David as co-author; review fixes and tests remain intact.

PR #302, opened 2026-08-13
refactor: move company profile content and tiers into the database

### LINKED ISSUE #280: [Refactor] Move company profile content and tiers into the database
### Description
Company roster (existence, tier, logo, website, order) lives in the DB, but everything richer (bio, socials, facts, interests, booth actions, tier styling) stays hardcoded in `src/edition/` (`DiamondCompanies.tsx`, `GoldCompanies.tsx`, `SilverCompanies.ts`, `BronzeCompanies.ts`, `actions.ts`), matched to the DB row by case-insensitive name string (`CompanyByName.ts`). New ranks/tiers beyond the fixed `DIAMOND`/`GOLD`/`SILVER`/`BRONZE` enum push this from "should move to DB" to "tiers themselves need to move to DB too."

### Why
The string-matching approach is already fragile in practice, not hypothetically: `edition/actions.ts` carries duplicate aliases for the same company because names drifted (`apr` + `"apr - technology solutions"`); two unsynced copies of `website` exist (DB vs. hardcoded `modalInformation.website`); a company with a DB row but no tier-file entry silently loses everything but a bare logo, with no error anywhere; and `company-tier-access.ts`/`tier-styling.ts` hardcode `Record<CompanyTier, ...>` maps keyed off the fixed 4-value enum, so a new rank means editing code, not data.

### How
Introduce `CompanyRank`/`CompanyRankStyle` (replacing the `Tier` enum), and `CompanyProfile`/`CompanyDisplayStyle` (1:1 companion tables for bio/socials/facts/video and logo sizing, keeping presentation fields out of `Company` itself), plus a plain `Company.interests`/`Interest.companies` m:n relation (implicit, no join table, mirroring the existing `User.interests` pattern). `socialLinks`/`facts` are JSONB — small, bounded/ordered blobs, no need to query by individual entry. Add `Action.companyId` FK, replacing `edition/actions.ts`'s `boothActions` name-matching map. Backfill from the 4 existing tier files using today's case-insensitive matching (the last place that logic runs), convert `bodyText` from JSX to plain text, extend the admin backoffice with CRUD + reorder for `CompanyRank` and forms for profile/style/social links/facts/interests, then delete the per-company content from `src/edition/` once verified.

### Dependencies
None.

david-fs-valente:
## Summary

Completes #280 in one PR (schema landed previously in #280's first PR; this finishes the cutover instead of splitting it into follow-ups, per the original PR's own note).

- Drops the fixed `Tier` enum; `Company.rank`/`CompanyRank` is now required and is the single source of truth for tier name, order, gradient styling, and whether a rank gets an internal `/company/[name]` page.
- Every consumer (public roster, company page, `CompaniesContainer`/`Company` link routing, admin backoffice) now reads rank/profile/display-style/interests from the DB instead of `src/edition/{Diamond,Gold,Silver,Bronze}Companies.{tsx,ts}` + `CompanyByName.ts`.
- Booth-action completion (`savedStudentService.saveStudent`) now resolves via `Action.companyId` instead of `edition/actions.ts`'s name-matching map (which carried known duplicate aliases, e.g. `"apr"` / `"apr - technology solutions"`).
- Admin backoffice gains full CRUD + drag-and-drop reorder for `CompanyRank` (replacing the fixed 4-lane tier board with a dynamic N-lane rank board), plus a company content form for profile/display-style/interests, plus a company picker on the Actions form/list.
- Deletes the now-dead `src/edition/` per-company content files and the `FactData`/`ModalProps` types that only existed to type them.

## Notable fixes found during this work

- Gradient styling moved from hardcoded Tailwind arbitrary-value classes to inline styles — Tailwind's JIT scanner can't pick up a `from-[${color}]` class built from runtime DB data, only literal strings in source.
- `company/[name]/page.tsx` now decodes the `name` param before querying — it can arrive with a literal `%20` undecoded, which the old `CompanyByName.ts` lookup handled but got dropped in an earlier draft of this cutover (caught via manual testing against a real local DB, not by the test suite).
- The internal-page gate now checks `rank.style.hasInternalPage` uniformly instead of only hard-blocking `SILVER` explicitly — closes a pre-existing gap where `BRONZE` wasn't blocked at the page level the same way, even though it was never linked to from the roster.

## Test plan

- [x] `pnpm typecheck`, `pnpm lint`, `pnpm test` (262 tests) all pass
- [x] `pnpm build` succeeds
- [x] Manually verified against a real local Supabase DB (`supabase start` + `prisma migrate deploy`, migrating a snapshot with real seeded company data through both #280 migrations):
  - `/api/companies/roster` returns companies correctly grouped by rank with style/interests/hasContent
  - Internal company pages render real bodyText (multi-paragraph split), facts (icon lookup), video embed, and social links; 404 for ranks without an internal page and for ranks with one but no profile content
  - Admin: CompanyRank list/create/reorder/delete (blocked while companies are assigned, allowed once empty), company edit page's content form writes profile/display-style/interests, action admin form assigns a company to a booth action


hugo2006alm (changes requested):
AGENTS.md:L117: medium: contributor guide still says tier/company content and booth mappings are hardcoded and directs changes to the removed `getBoothActionName()` flow. Update this section to document DB-backed `CompanyRank`/profile content and `Action.companyId`; otherwise future changes will follow an architecture this PR deletes.

src/components/CompanyContentForm/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+            },
+          ],
+        },
+        {
+          kind: "fields",
+          title: "Logótipo",
+          fields: [
+            { kind: "number", name: "logoWidth", label: "Largura" },
+            { kind: "number", name: "logoHeight", label: "Altura" },
+            { kind: "text", name: "className", label: "Classe CSS" },
```
dinisjcorreia: [P2] This field accepts arbitrary Tailwind classes from the database, but Tailwind generates CSS at build time by scanning source literals. The migrated values happen to exist in the SQL source, but any new value entered by an admin after deployment can have no generated CSS and silently do nothing. Use a bounded/safelisted choice, or store a numeric display width and apply it as an inline style.

src/components/Companies/CompanyPageSection/index.tsx
```diff
@@ -52,49 +43,49 @@ const CompanyPageSection: React.FC<CompanyPageSectionProps> = ({
           )}
           <div className="flex flex-col gap-y-2 px-4 text-center">
             <p className="text-3xl font-bold text-white md:text-5xl">
-              <span>{modalInformation.title}</span>
+              <span>{company.name}</span>
```
dinisjcorreia, line 46: [P2] This replaces `modalInformation.title`, but several migrated titles differ from `Company.name` (`armis` -> `Armis`, `hitachi` -> `Hitachi Solutions`, `deloitte` -> `Deloitte`, `Devscope` -> `DevScope`). Since the migration neither stores those titles nor normalizes the company names, existing public profile headings regress after deploy. Preserve a display title or backfill an equivalent canonical display name.

prisma/migrations/20260809100000_add_company_profile_and_rank/migration.sql
```diff
@@ -0,0 +1,375 @@
...
+-- src/edition/{Diamond,Gold,Silver,Bronze}Companies.{tsx,ts} (matched by
+-- name, same case-insensitive-in-practice matching CompanyByName.ts did -
+-- the values below use the exact Company.name rows already seeded by
+-- prisma/migrations/20260714011113_company_display_fields).
+INSERT INTO "CompanyDisplayStyle" ("id", "companyId", "logoWidth", "logoHeight", "className", "updatedAt")
+SELECT gen_random_uuid(), c."id", v."logoWidth", v."logoHeight", v."className", CURRENT_TIMESTAMP
+FROM (
+  VALUES
+    ('APR - Technology Solutions', 4215, 4215, 'w-42'),
+    ('armis', 149, 149, NULL),
```
dinisjcorreia, line 157: [P2] The deleted `DiamondCompanies.tsx` gives Armis `className: "w-42"`, but this backfill writes `NULL`. Saving the migrated data therefore changes the existing logo size. Backfill `w-42` here (or preserve the equivalent display width if `className` is replaced).

dinisjcorreia (changes requested):
Found three user-visible regressions in the database cutover. CI passes on this head, but these should be fixed before merge.

dinisjcorreia (approved):
Re-reviewed d5836ae. All three requested fixes are addressed: logo-width choices are bounded and generated, profile headings are restored, and the Armis backfill is repaired with a follow-up migration. Tests, typecheck, lint, and current CI pass.

PR #305, opened 2026-08-16
fix: stabilize account deletion and admin ordering

### LINKED ISSUE #285: [Fix] Close orphaned-account gap between Supabase Auth and public.User on delete

### LINKED ISSUE #287: [Test] Validate rate limiter behavior under realistic event load
### Description
Validate whether the current authenticated-user fixed-window rate limiter behaves acceptably under realistic Fallstack event traffic before deciding whether a more complex limiter or shared store is justified.

### Why
`createRateLimiter` (`src/lib/rateLimit.ts`) is a single-server, in-memory, fixed-window limiter by design: a key can use its full request budget on both sides of a window boundary, and there is no shared state across replicas. Those are known, accepted trade-offs today, but they have never been validated against the concurrency and traffic shape Fallstack's QR/upload/event endpoints will actually see on event day. This is the one issue before the release-readiness gate (#344) where evidence, not assumption, decides whether the current limiter is good enough or needs to become a shared token-bucket implementation before the event.

### Scope
- Use the existing k6 harness rather than creating a new load-testing stack.
- Exercise representative QR/upload-ticket/event endpoints with realistic concurrent authenticated users.
- Include bursts around window boundaries and verify expected `429` behavior.
- Measure whether normal event traffic is incorrectly throttled or whether bursts create an operational problem.
- Verify behavior with the deployment topology expected for the event.
- Only replace the current fixed-window implementation with token-bucket/shared-state logic if test evidence demonstrates a real problem.
- Record the tested concurrency assumptions and results for the final release-readiness gate #344.

### Out of scope
The previous dormant `getClientIp` concern is removed from this issue because the active limiter is keyed by authenticated user and that dead-code concern is not the event-readiness question being validated here.

### Acceptance Criteria
- k6 tests exercise representative QR/upload-ticket/event endpoints with realistic concurrent authenticated users, including bursts at rate-limit window boundaries.
- Expected `429` behavior at window boundaries is verified against actual results.
- The results state explicitly whether normal event traffic is incorrectly throttled or whether bursts create an operational problem.
- Results are captured against the deployment topology expected for the event, not a synthetic/local topology.
- The fixed-window limiter is only replaced with token-bucket/shared-state logic if test evidence demonstrates a real problem — no replacement happens without that evidence.
- The tested concurrency assumptions and results are recorded in a form #344 can consume.

### Dependencies
- Blocks #342 — staging readiness automation incorporates this issue's rate-limiter validation.
- Blocks #344 — the release-readiness gate records this issue's tested concurrency assumptions and results.

### LINKED ISSUE #288: [Fix] Add unique constraints to close TOCTOU gap in admin reordering
### Description
Three DB-backed admin reordering features — Schedule (`scheduleService.ts`), FAQ (`faqService.ts`), and the Company tier-board (`companyService.ts`'s `updateCompanyTierBoard`) — share a TOCTOU-shaped concurrency gap: a write plan gets computed from a read that isn't fully isolated from a concurrent second admin doing the same thing.

### Why
This isn't classic TOCTOU-as-vulnerability — every route involved is gated behind `auth: "admin"`, so only an already-fully-trusted admin can reach it; the race is between two equally-privileged actors, not an attacker vs. a boundary. The worst-case outcome is purely presentational (overlapping timetable slots, two FAQs sharing a display order) — recoverable by re-editing, not a breach of anything security-critical. Given low probability (a handful of trusted admins, narrow concurrent-edit window) and low severity, a DB unique constraint is the right-sized fix: cheap, and it converts "silently produces two rows sharing a slot" into a loud, visible constraint-violation error.

### How
Migration: add a unique constraint on `(day, order)` for `ScheduleEvent` and `(order)` for `FaqEntry`. Handle the resulting constraint-violation gracefully in the admin UI/API (catch it and surface "someone else already used this slot — please retry" rather than a raw 500). Add a short code comment near `scheduleService.ts`/`faqService.ts` explaining why a unique constraint is sufficient here rather than row-locking/serializable transactions. Leave the Company tier-board's lost-update risk as a documented, accepted risk under the same reasoning — revisit with optimistic concurrency only if usage patterns change.

### Dependencies
None.

### LINKED ISSUE #304: [Security] Patch high-risk production dependency findings
### Description
Production dependency audit reports patched releases available for transitive `fast-uri`, `brace-expansion`, and `nanoid` dependencies used by the Sentry build toolchain.

### Scope
- Refresh the frozen lockfile/overrides to remove all high/critical production findings.
- Re-run the complete quality gate after the dependency changes.
- Confirm the production audit no longer reports the targeted high/critical findings.
- Avoid bundling unrelated dependency upgrades into the same fix unless required to resolve the findings.

### Planning
- **Sprint 3 — 24 Aug to 7 Sep**
- **Priority: Urgent**
- Security/release blocker for the 2026 edition.

hugo2006alm:
## Stabilization summary

| Issue/finding | Original problem | Root cause | Fix | Tests added | Migration/deployment impact | Manual action |
|---|---|---|---|---|---|---|
| #285 / PR #296 | Deleting app accounts left Supabase Auth orphans; failed cleanup could leave either side unusable | No cross-schema FK and one-sided delete paths | Central Auth-first deletion, fail-closed placeholder cleanup, cascade action completions, read-only orphan audit | Auth deletion ordering/failure/race tests; cascade migration check | Run Prisma migration before app rollout | Run `supabase/audit-orphaned-accounts.sql` in every environment and investigate results |
| #287 | Dormant proxy-IP trust code and unverified fixed-window behavior | Active routes use authenticated user IDs; IP parser had no callers and depended on topology | Removed dormant IP parser; documented fixed-window/single-replica boundary and staging k6 gate | Retained limiter budget/reset/isolation tests | None | Run QR/upload k6 scenarios against staging before event |
| #288 | Concurrent admin reorder could persist duplicate schedule/FAQ positions | Read-then-write plans lacked DB uniqueness | Deterministic normalization migration; deferred unique constraints; retryable HTTP 409 conflicts | Schedule/FAQ conflict regression tests | Migration normalizes current positions, then adds constraints | Back up DB; inspect normalized schedule/FAQ order after migration |
| #304 | High-risk transitive dependency findings | Frozen graph contained vulnerable patched dependencies | Same-major overrides and refreshed lockfile | Full 296-test suite; production and full audit show no known findings | Rebuild application image | None |

## Authorship

#285 preserves Dinis Correia's PR #296 work and David Valente's fail-closed correction in one logical commit with original/co-author attribution. PR #296 can be superseded after review.

## Validation

- `pnpm exec vitest run --maxWorkers=2`: 69 files, 296 tests
- `pnpm typecheck`: pass
- `pnpm lint`: 0 errors (Windows checkout emits pre-existing CRLF warnings; Linux CI authoritative)
- `pnpm audit --prod --audit-level high`: no known vulnerabilities
- `pnpm audit --audit-level high`: no known vulnerabilities
- Gitleaks over PR commits: no leaks
- Local Next production compilation/static generation completed; Windows standalone symlink creation is blocked by EPERM. Linux CI/Docker jobs are authoritative.

Closes #285
Closes #287
Closes #288
Closes #304

## Dependency and documentation follow-up

- Updated PostCSS to 8.5.26 and constrained routine Dependabot PRs to patch/minor updates targeting `dev`.
- Updated `AGENTS.md` with the exact CI/CD gates, frozen-install expectations, deployment flow, and test-first/TDD guidance.


src/application/services/authApplicationService.ts
```diff
@@ -181,7 +205,35 @@ export async function completeOAuthSignIn(input: {
     // Unconfirmed - discard the dangling placeholder instead of relinking
     // onto it, so AuthNEI (institutionally-verified) can claim the email
     // fresh below rather than inheriting potentially attacker-planted data.
-    await deleteUser(existingByEmail.id);
+    try {
+      await deleteUserAccount(existingByEmail.id);
```
dinisjcorreia: [P1] Do not delete when confirmation status is unknown

`isEmailConfirmed()` returns `false` for every `getUserById()` error, not only for a successful lookup with no `email_confirmed_at`. This new call therefore deletes the Supabase identity and cascaded app profile for a legitimately confirmed legacy user whenever that lookup transiently fails; the added `fails closed` test currently codifies this destructive path. Throw/abort on lookup errors and invoke deletion only after a successful lookup explicitly proves the identity is unconfirmed.

pnpm-workspace.yaml
```diff
@@ -1,16 +1,25 @@
...
+  '@typescript-eslint/type-utils': ^8.8.1
+  '@typescript-eslint/utils': ^8.8.1
+  brace-expansion@5.0.7: 5.0.9
+  brace-expansion@5.0.8: 5.0.9
+  brace-expansion@<1.1.18: 1.1.18
+  brace-expansion@>=2.0.0 <2.1.4: 2.1.4
+  brace-expansion@>=4.0.0 <5.0.9: 5.0.9
+  fast-uri@>=3.0.0 <3.1.5: '>=3.1.5'
+  js-yaml@>=4.0.0 <4.3.1: '>=4.3.1'
+  nanoid@<3.3.18: '>=3.3.18'
```
dinisjcorreia: [P2] Keep security overrides inside parent major ranges

These open-ended replacement specs cross dependents' declared majors: the lockfile installs `fast-uri@4.1.2` for Ajv's `^3.0.1`, `js-yaml@5.3.0` for `@eslint/eslintrc`'s `^4`, and `nanoid@6.0.1` for PostCSS's `^3.3.17`. Current CI passing does not make unsupported transitive majors safe, and a later lock refresh can jump again. Pin the patched compatible releases (`3.1.5`, `4.3.1`, and `3.3.18`) or upgrade each parent with the migration plan required by this repo's policy.

dinisjcorreia (changes requested):
Blocking findings below. Targeted auth/order/migration tests pass (44/44), production audit is clean, and Prisma schema validates. GitHub also currently reports this branch conflicts with `dev`.

src/application/services/authApplicationService.ts
hugo2006alm: Fixed. A failed/unknown `getUserById()` lookup now aborts the OAuth flow and does not delete either the Auth identity or the application row. The regression test was updated accordingly.

pnpm-workspace.yaml
hugo2006alm: Fixed. The three overrides are now pinned to the patched compatible releases: `fast-uri` 3.1.5, `js-yaml` 4.3.1, and `nanoid` 3.3.18.

PR #320, opened 2026-08-18
fix(auth): request NEI Global role scopes

hugo2006alm:
## Summary
- add optional `NEXT_PUBLIC_AUTHNEI_GLOBAL_PROJECT_ID` configuration and wire it through Next.js Docker build/runtime configuration
- keep AuthNEI's existing `openid email profile` scopes and, when the global project id is configured, also request:
  - `urn:zitadel:iam:org:projects:roles`
  - `urn:zitadel:iam:org:project:id:<NEI_GLOBAL_PROJECT_ID>:aud`
- keep the project id optional so existing deployments continue to authenticate before the shared ZITADEL project is configured
- update `.env.example` to reflect the Compose/Coolify-derived app URL and document the global project id while preserving dev-only staging E2E settings

## Coolify
This builds on #318, which already removed the dev Compose `.env` dependency, switched the public base URL to `SERVICE_URL_WEB` + `/api` with a local fallback, and let `DIRECT_URL` inherit `DATABASE_URL` when appropriate.

`SERVICE_FQDN_WEB` is not used because Fallstack currently has no setting that consumes only a hostname; the relevant app setting consumes a full URL. No `COOLIFY_URL` dependency is introduced.

## AuthNEI / ZITADEL
Supabase custom OIDC treats scopes supplied by `signInWithOAuth` as the requested scope set, so the app explicitly retains `openid email profile` instead of supplying only the ZITADEL role scopes.

Set `NEXT_PUBLIC_AUTHNEI_GLOBAL_PROJECT_ID` in Coolify to the Project ID of the shared `NEI Global` project. If left empty, AuthNEI login retains its current scope behavior.

This PR only requests the shared role claims. Fallstack's current admin authorization remains its existing Prisma `adminRole` model; mapping the ZITADEL `admin` role into that authorization model is intentionally not silently changed here.


hugo2006alm:
Superseded by #305. Fallstack now authenticates directly against AuthNEI/ZITADEL instead of requesting ZITADEL scopes through Supabase custom OIDC, so this PR's approach is no longer used. The required NEI Global audience/scopes are included in the direct ZITADEL configuration merged through #305.

PR #329, opened 2026-08-20
fix(config): simplify Coolify environment setup

hugo2006alm:
## Summary

Align Fallstack deployment configuration with the Coolify env conventions used across the other NEI apps.

### Changes
- derive the public app/API URLs from `SERVICE_URL_WEB`
- generate `JWT_SECRET` and `AUTH_SECRET` through Coolify `SERVICE_REALBASE64_64_*` magic variables when not explicitly configured
- default `AUTH_ISSUER_URL` to `https://auth.nei-isep.org`
- derive ZITADEL scopes and role-claim names from `AUTH_PROJECT_ID` / `AUTH_GLOBAL_PROJECT_ID`
- request both Fallstack and global project audiences by default
- remove unused `offline_access` from the custom OIDC flow
- keep external DB, Supabase and ZITADEL application credentials mandatory
- remove duplicate Compose `PORT` / `HOSTNAME` configuration and host port publishing; the runner keeps its internal port in the Dockerfile
- align the direct-build `NEXT_PUBLIC_BASE_URL` fallback with Fallstack's port `4000`
- document the resulting local-vs-Coolify behavior in `.env.example`

## Coolify values that still need manual configuration

- `DATABASE_URL` (`DIRECT_URL` only if different)
- `NEXT_PUBLIC_SUPABASE_URL`
- `NEXT_PUBLIC_SUPABASE_ANON_KEY`
- `SUPABASE_SERVICE_ROLE_KEY`
- `AUTH_PROJECT_ID`
- `AUTH_GLOBAL_PROJECT_ID`
- `AUTH_CLIENT_ID`
- `AUTH_CLIENT_SECRET`
- `ZITADEL_ORG_ID`
- `ZITADEL_ROLE_ASSIGNER_TOKEN`

Observability values remain optional.

This PR intentionally does not change the Gitleaks/security workflow issue discussed separately.

### DIFF
--- a/src/config/env.client.ts
+++ b/src/config/env.client.ts
@@ -6,17 +6,11 @@ import { z } from "zod";
 const clientEnvSchema = z.object({
   NEXT_PUBLIC_SUPABASE_URL: z.string().url(),
   NEXT_PUBLIC_SUPABASE_ANON_KEY: z.string().min(1),
-  // Environment-related fix, not strictly in scope of the Docker non-root
-  // change this shipped alongside (#218/#207): the Dockerfile's
-  // `ARG NEXT_PUBLIC_BASE_URL=""` default means "unset" the same way
-  // `NEXT_PUBLIC_SENTRY_DSN` below does, but without this preprocess step
-  // the empty string reached `.url()` directly and failed validation
-  // instead of falling through to `.default()` — so a build run without
-  // this build arg supplied (e.g. `docker build` with no `--build-arg`s)
-  // failed at `next build` time instead of using the documented default.
+  // Explicit empty values still mean "unset" so direct builds can fall back
+  // to the same local port used by the Docker runner and .env.example.
   NEXT_PUBLIC_BASE_URL: z.preprocess(
     (value) => (value === "" ? undefined : value),
-    z.string().url().default("http://localhost:3000/api")
+    z.string().url().default("http://localhost:4000/api")
   ),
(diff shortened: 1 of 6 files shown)


src/config/env.client.ts
```diff
@@ -6,17 +6,11 @@ import { z } from "zod";
...
-  // the empty string reached `.url()` directly and failed validation
-  // instead of falling through to `.default()` — so a build run without
-  // this build arg supplied (e.g. `docker build` with no `--build-arg`s)
-  // failed at `next build` time instead of using the documented default.
+  // Explicit empty values still mean "unset" so direct builds can fall back
+  // to the same local port used by the Docker runner and .env.example.
   NEXT_PUBLIC_BASE_URL: z.preprocess(
     (value) => (value === "" ? undefined : value),
-    z.string().url().default("http://localhost:3000/api")
+    z.string().url().default("http://localhost:4000/api")
```
dinisjcorreia: [P2] Keep the native dev fallback on port 3000

This schema is shared by native development, where pnpm dev still runs plain next dev on port 3000. When NEXT_PUBLIC_BASE_URL is unset or empty, the new fallback sends every httpClient request to localhost:4000 instead. Dockerfile and Compose now inject their own port-4000 values, so they do not require changing this shared fallback; keep it at 3000 (and align the local example), or explicitly move the dev server to 4000.

dinisjcorreia (changes requested):
One blocking local-development regression. Compose expansion, generated secrets, and CI checks otherwise look correct.

PR #335, opened 2026-08-21
feat: add DB-backed dynamic translations

### LINKED ISSUE #156: [Feat] Add i18n / string externalization for user-facing text
### Description
Externalize Fallstack's user-facing static text and add English alongside Portuguese for the 2026 edition. Portuguese remains the default/base language and the committed scope is PT+EN.

### Scope
1. Adopt `next-intl` for Server and Client Components.
2. Decide and document locale resolution/routing (URL-based vs cookie/header) with a visible language switcher.
3. Externalize shared/navigation UI first, then page-level copy incrementally.
4. Add PT and EN message files with human review for event/NEI terminology.
5. Define fallback behavior and locale persistence.
6. Stop introducing new hardcoded Portuguese-only user-facing strings once the i18n foundation lands.
7. Add regression coverage for locale switching and representative critical pages.

### Coordination with #279
#279 is already in review and no longer waits for this issue. Align the static i18n locale identifiers/fallback rules with the DB-backed translation model from #279 so Fallstack has one language convention.

### Dependencies
No remaining blocker. The previous accessibility prerequisite #141 is completed.

### Planning
- **Sprint 4 — 7 Sep to 21 Sep**
- **Priority: High**

### LINKED ISSUE #279: [Feat] Add DB-backed dynamic text translation
### Description
Static-text i18n (#156) covers hardcoded UI strings, while DB-backed text such as `FaqEntry.question`/`answer`, `ScheduleEvent.activity`, and `Interest.name` needs its own persistence/localization mechanism.

Implementation is currently in review in **PR #335**.

### Scope
- Store the relevant DB-backed fields as PT/EN translations.
- Keep translation validation at the repository/domain boundary.
- Preserve existing data through a safe migration/backfill.
- Update admin CRUD and public reads to use localized values.
- Keep Sponsor brand identifiers un-translated unless a real user-facing translatable field requires it.
- Preserve stable IDs/relations when DB uniqueness on translated JSON cannot be expressed directly.

### Coordination with #156
#279 is already underway and is **not blocked** on #156 anymore. The two implementations must, however, converge on the same locale identifiers and fallback behavior so static and DB-backed translations do not create separate language conventions.

### Implementation
- PR #335
- The PR should close this issue when merged (`Closes #279`).

### Planning
- **Sprint 3 — 24 Aug to 7 Sep**
- **Priority: High**
- **Status target: In Review**

dinisjcorreia:
## Summary
- migrate FAQ question/answer, schedule activity, and interest name fields from text to JSONB translations
- add PT/EN translation value object with repository-boundary validation and PT fallback for migrated rows
- update admin CRUD forms and public DTOs/routes to write and select localized values
- preserve interest relations through stable IDs after dropping name uniqueness
- keep translated admin search and PT sorting behavior
- route Prisma files through Prisma formatter in lint-staged

Sponsor fields remain unchanged: sponsor name is a brand identifier; logo and website are not translatable text.

## Migration
Includes `prisma/migrations/20260821120000_add_dynamic_translations/migration.sql`. It backfills existing values as `{"PT": "..."}` and drops FAQ-question/interest-name uniqueness. Migration was not applied locally because local Postgres was offline.

## Verification
- `pnpm lint` (passes; existing warnings remain)
- `pnpm test` (289 passed)
- `pnpm typecheck`
- `pnpm exec prisma validate`
- `pnpm build`
- `docker build --target migrator -t fallstack-migrator-check .`
- migrator image runs as `nextjs`
- dev server `/` returned 200; DB-backed live routes could not be exercised without local Postgres

Skipped optional case-insensitive soft duplicate check; trusted-admin CRUD and accepted soft-dedup tradeoff cover current scale.

## Issues
Closes #279


prisma/migrations/20260821120000_add_dynamic_translations/migration.sql
```diff
@@ -0,0 +1,12 @@
+DROP INDEX "Interest_name_key";
```
david-fs-valente, line 1: **Blocking.** This drops the unique indexes on `Interest.name` and `FaqEntry.question` with nothing replacing them at the app layer — `interestService.createInterestForAdmin` has no duplicate check, so two identical interests/FAQ questions can now be created silently. Suggest adding expression unique indexes on the JSONB path, e.g. `CREATE UNIQUE INDEX ON "Interest" ((name->>'PT'))` (and one for `EN`), then mapping the resulting `P2002` to a friendly error the same way `faqService.ts`'s `faqConflict()` already does for `FaqEntry.order`.

src/domain/i18n/translations.ts
```diff
@@ -0,0 +1,78 @@
...
+export enum Language {
+  PT = "PT",
+  EN = "EN",
+}
+
+export type TranslationValues = Record<Language, string>;
+
+export const translationsSchema = z
+  .object({
+    [Language.PT]: z.string().min(1),
```
david-fs-valente: **Blocking.** `PT` requires `.min(1)`, but the migration wraps every legacy string as-is with no length check, and nothing pre-existing enforced non-empty strings. If any current `Interest.name`/`FaqEntry.question`/`ScheduleEvent.activity` row is `''`, every read (`parseInterest`/`parseFaqEntry`/`parseScheduleEvent`, called in an unguarded `.map()`) throws and takes the whole list endpoint down, not just that row. Please confirm no empty-string rows exist in prod, and consider making the per-row parse resilient to one bad row regardless.

src/app/api/faqs/route.ts
```diff
@@ -3,11 +3,15 @@ import { NextResponse } from "next/server";
...
 import { getFaqEntries } from "@/application/services/faqService";
+import { resolveLanguage } from "@/domain/i18n/translations";
 
 export const GET = defineHandler({
   auth: "public",
-  handler: async () => {
+  handler: async ({ req }) => {
     const faqs = await getFaqEntries();
-    return NextResponse.json(faqs.map(toFaqDto));
+    const language = resolveLanguage(
```
david-fs-valente: **Non-blocking.** `?lang=` is honored here (and in `schedule`/`companies/roster`) but ignored by the page components, which only read `Accept-Language`. Since a language picker is planned, worth wiring `?lang=` through the pages too (via `searchParams`) and extracting this line into one shared `resolveRequestLanguage()` helper so every route/page resolves language the same way before the picker lands.

src/application/repositories/interestRepository.ts
```diff
@@ -27,41 +38,69 @@ export interface AdminInterestQuery {
...
+    await prisma.interest.findMany({
+      where: interestWhere(search),
+      select: { id: true, name: true, _count: { select: { users: true } } },
+    })
+  )
+    .map(parseInterest)
+    .sort((a, b) =>
+      a.name.PT.localeCompare(b.name.PT, "pt", { sensitivity: "base" })
+    );
+  if (order === "desc" && sort === "name") interests.reverse();
```
david-fs-valente: **Non-blocking.** `sort === "name"` is always true now that `ADMIN_SORTABLE_FIELDS` was removed from this file and `name` is the only sortable field — `sort` is otherwise unused here. Consider `if (order === "desc") interests.reverse();` so the condition doesn't imply `sort` still gates anything.

src/application/services/interestService.ts
```diff
@@ -25,11 +26,14 @@ export async function listInterestsForAdmin(query: AdminInterestQuery) {
   return { items, totalCount };
```
david-fs-valente: **Non-blocking.** `listInterestsForAdmin` above runs `countInterestsForAdmin` as a separate full COUNT even though `findInterestsForAdmin` already fetches every matching row unpaginated. Can just use `items.length` and drop the second query.

src/application/repositories/interestRepository.ts
```diff
@@ -27,41 +38,69 @@ export interface AdminInterestQuery {
...
+  prisma.interest
+    .findUnique({ where: { id } })
+    .then((interest) => (interest ? parseInterest(interest) : null));
 
 export const countInterestUsers = (id: string) =>
   prisma.user.count({ where: { interests: { some: { id } } } });
 
-export const createInterest = (name: string) =>
-  prisma.interest.create({ data: { name } });
+const toJson = (value: TranslationValues) =>
```
david-fs-valente: **Improvement.** This `toJson()` is copy-pasted identically in `faqRepository.ts` and `scheduleRepository.ts`. Worth exporting once from `src/domain/i18n/translations.ts`.

src/application/repositories/companyRepository.ts
```diff
@@ -2,8 +2,24 @@ import "server-only";
 
 import { Prisma } from "@prisma/client";
 
+import { Translations } from "@/domain/i18n/translations";
+
 import prisma, { DbClient } from "./database";
 
+const parseInterest = <T extends { name: Prisma.JsonValue }>(interest: T) => ({
```
david-fs-valente: **Improvement.** This `parseInterest` is reimplemented 4x (here, `interestRepository.ts`, `studentRepository.ts`, and inlined with no name in `savedStudentRepository.ts`). Consider a generic `parseTranslatedField(entity, field)` in `translations.ts` that each repo calls with its own field name (`"name"`, `"question"`, `"activity"`) instead of one bespoke copy per entity.

src/application/repositories/faqRepository.ts
```diff
@@ -46,7 +54,15 @@ export interface AdminFaqQuery {
 
 function faqWhere(search?: string) {
```
david-fs-valente: **Improvement.** This JSON-path OR-filter (`faqWhere`) is hand-duplicated in `interestRepository.ts`/`scheduleRepository.ts`, differing only by field name. A shared `translatedFieldWhere(field, search)` helper would remove the triplication.

src/domain/i18n/translations.ts
```diff
@@ -0,0 +1,78 @@
...
+            : 0,
+        index,
+      };
+    })
+    .filter(
+      (preference): preference is typeof preference & { language: Language } =>
+        preference.language !== undefined && preference.quality > 0
+    )
+    .sort(
+      (left, right) => right.quality - left.quality || left.index - right.index
```
david-fs-valente: **Improvement.** `Array.prototype.sort` has been spec-stable since ES2019, so the `index` field and `|| left.index - right.index` tie-break just re-derive a guarantee already in place — equal-quality entries already keep their original header order. Can simplify to `.sort((left, right) => right.quality - left.quality)` and drop `index`.

src/application/dto/interestDto.ts
```diff
@@ -1,23 +1,39 @@
...
+  name: TranslationValues;
+}
+
+export const toInterestDto = (
+  interest: InterestEntity,
+  language: Language = Language.PT
+): InterestDto => ({
   id: interest.id,
-  name: interest.name,
+  name: Translations.fromJSON(interest.name).get(language),
```
david-fs-valente: **Improvement.** `Translations.fromJSON` re-parses data the repository's `parseInterest` already validated — same pattern in `toFaqDto`/`toCompanyRosterDto`/`toCompanyDisplayDto`. Since the repository already guarantees the `{PT,EN}` shape, this could index `interest.name[language]` directly and skip the redundant second Zod pass on every read.

david-fs-valente (changes requested):
Reviewed the JSONB translations migration. Two blocking issues (dropped uniqueness, possible crash on legacy empty-string rows), a few non-blocking inconsistencies, and several duplication/efficiency improvements. Details inline.

src/app/api/faqs/route.ts
dinisjcorreia: Fixed in 6aa998e: added shared resolveRequestLanguage(), wired ?lang= through all translated server pages, and passed resolved language into homepage FAQ/schedule/company API requests.

src/application/repositories/companyRepository.ts
dinisjcorreia: Fixed in 6aa998e: added shared parseTranslatedField(s) and reused it across interest, company, student, saved-student, FAQ, and schedule repositories.

src/domain/i18n/translations.ts
dinisjcorreia: Fixed in 6aa998e: persistence schema now accepts legacy empty strings, while admin request schemas still require non-empty PT and EN. Added regression coverage proving a migrated empty PT value reads without crashing.

src/application/repositories/interestRepository.ts
dinisjcorreia: Fixed in 6aa998e: removed the obsolete sort === "name" condition; descending order now reverses directly.

prisma/migrations/20260821120000_add_dynamic_translations/migration.sql
dinisjcorreia: Fixed in 6aa998e: migration now adds PT/EN JSONB expression unique indexes for Interest.name and FaqEntry.question. Interest and FAQ services map P2002 to friendly 409 responses.

src/application/repositories/interestRepository.ts
dinisjcorreia: Fixed in 6aa998e: translation serialization now lives once in toTranslationJson() under the i18n domain module.

src/application/dto/interestDto.ts
dinisjcorreia: Fixed in 6aa998e: DTOs and services now index repository-validated TranslationValues directly, avoiding a second Zod parse per read.

src/application/repositories/faqRepository.ts
dinisjcorreia: Fixed in 6aa998e: extracted translatedFieldWhere() in the repository layer and reused it for FAQ, interest, and schedule JSON-path search.

src/domain/i18n/translations.ts
dinisjcorreia: Fixed in 6aa998e: removed the redundant index tie-break and now relies on stable Array.sort for equal-quality language preferences.

src/application/services/interestService.ts
dinisjcorreia: Fixed in 6aa998e: interest repository returns its already-fetched matching rows; service derives totalCount from length and slices locally. Separate COUNT removed.

david-fs-valente (approved):

PR #357, opened 2026-09-09
test(e2e): expand event coverage across roles

### LINKED ISSUE #341: [Test] Expand event E2E coverage to employee and admin critical flows
### Description
Expand the existing Playwright staging coverage beyond the current student-oriented flow so the event can be exercised end-to-end across the roles that operate Fallstack on the day.

### Scope
- Add a synthetic authenticated Employee flow covering a representative event action/scan workflow.
- Add duplicate/retry coverage for event actions that must not be applied twice.
- Add a representative Admin flow against staging.
- Verify role boundaries between Student, Employee, Admin and Super Admin where relevant.
- Exercise the critical path using real staging HTTP/database state, not only mocked browser state.
- Keep destructive actions isolated to synthetic staging data/accounts.

### Acceptance Criteria
- A staging run proves one representative Student -> Employee interaction end-to-end.
- Duplicate/retry behavior is asserted.
- At least one Admin operation is covered.
- Unauthorized role transitions/actions are rejected.
- Tests are documented and safe to rerun against staging.

dinisjcorreia:
## Summary
- preserve configured Admin vs Super Admin tier while retaining AuthNEI global-admin gating
- cover Student QR to Employee scan persistence and exact retry rejection against staging state
- cover isolated Admin FAQ create/read/delete and cross-role authorization boundaries
- document synthetic account setup, rerun behavior, cleanup, and required storage states

## Verification
- pnpm lint
- pnpm typecheck
- pnpm exec vitest run --maxWorkers=2 (313 passed)
- production build with CI Node 20.19.1
- pnpm exec prisma validate
- Playwright harness locally (3 staging tests discovered and safely skipped without credentials)

Live staging execution requires the four synthetic account storage-state files, which are intentionally not available in this workspace.

Closes #341


tests/e2e/playwright/event-role-flows.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      const code = studentSession.student!.code;
+
+      const qrResponse = await student.get("/api/qrcode");
+      expect(qrResponse.status()).toBe(200);
+      const { data: token } = (await qrResponse.json()) as { data: string };
+
+      const firstScan = await employee.post("/api/saved", {
+        data: { token, comment: marker },
+      });
+      expect([201, 409]).toContain(firstScan.status());
```
david-fs-valente, line 67: `expect([201, 409]).toContain(firstScan.status())` tolerates reruns against the same synthetic student/company relation, but it also means a real regression in the create path (e.g. always returning 409) could hide behind the already-saved branch and this assertion would still pass. Consider using a fresh/clean synthetic account per run so the first-scan assertion is a deterministic 201 instead.

david-fs-valente (changes requested):
Verified locally: checked out this branch, ran the touched unit tests (16/16 passed), `pnpm typecheck` (clean), and lint on new/changed files (clean). Every HTTP status/body/field assumption in the new Playwright spec was independently cross-checked against the real route/service source and matches exactly. This plausibly satisfies issue #341's acceptance criteria. Requesting changes on two items before merge:

1. **Branch naming**: the branch is `test/expand-event-e2e`, but AGENTS.md's Conventional Branch list only allows `feature/`, `feat/`, `bugfix/`, `fix/`, `hotfix/`, `release/`, `docs/`, `chore/` — there is no `test/` type. Since this PR also ships a real behavior change (`resolveAdminRole` in `src/domain/auth/authPolicy.ts`, changing how admin tiers are resolved in `sessionService.ts`/`userRepository.ts`), `fix/` would have been the compliant prefix. Please rename the branch (or note why not) rather than merging a non-compliant branch name.
2. **Unverified in practice**: the PR description states the three new Playwright tests were discovered and skipped without staging credentials, so none of them have actually been run against live staging. The status codes/bodies they assert on all match the real handlers, but the tests' actual pass/fail behavior is unverified. Please have someone with staging credentials run them before merging, per AGENTS.md's "don't claim success without observing it" rule.

tests/e2e/playwright/event-role-flows.spec.ts
dinisjcorreia: Valid. Fixed in 1aca309 on replacement PR #366: the test now reads the employee company history first, requires at most one existing Student/company relation, then asserts exact 201 for a clean pair or exact 409 for a safe rerun. This keeps persistent synthetic accounts rerunnable without accepting an unexplained status.

dinisjcorreia:
Both requested changes were valid.

- Renamed branch to `fix/expand-event-e2e`. GitHub closed #357 when the remote branch was renamed, so replacement PR #366 preserves the commits and compliant branch.
- Tightened first-scan assertion in `1aca309`: exact `201` when staging pre-state is clean; exact `409` only for a known persistent rerun.
- Full local pre-push hook passed: lint, typecheck, 77 test files / 313 tests.

Live staging execution remains unverified here because no synthetic storage-state credentials are available. PR #366 states this explicitly; someone holding those staging sessions must run it before merge. Please re-review #366.

Issue #340, opened 2026-08-23
[Infra] Set up Google Wallet Generic Pass issuer and publishing access


hugo2006alm:
### Description
Prepare the external Google Wallet infrastructure required before the Fallstack attendee identification pass can ship as a **Generic Pass**.

The Wallet pass is not an event admission ticket. It is an additional way for a student to present the existing Fallstack QR/barcode used by companies to access the student's authorized information through the current website flow.

### Scope
- Create or confirm the Google Wallet Issuer account for the organization.
- Create/select the Google Cloud project used by the Fallstack Wallet integration.
- Enable the Google Wallet API.
- Create the service account and minimum required credentials/permissions.
- Create a test **Generic Class** for the Fallstack 2026 attendee/student pass.
- Generate at least one demo **Generic Object** containing a representative QR/barcode value and verify `Add to Google Wallet` end-to-end.
- Keep the existing Fallstack QR/backend flow as the source of truth; Wallet only presents the identifier.
- Request production/publishing access early enough to absorb external approval delays.
- Document which values are environment-specific and where secrets are configured.

### Acceptance Criteria
- A demo Fallstack attendee identification pass can be added to Google Wallet using non-production configuration.
- The demo Generic Object renders a QR/barcode compatible with the existing Fallstack identification flow.
- Publishing access has been requested (and tracked if still pending).
- Dev/staging/prod configuration requirements are documented without committing credentials.
- #3 can be implemented without waiting on basic issuer setup.

### Dependencies
- Blocks #3 — #3 needs this issuer/service-account/Generic Class setup before implementation can start.

david-fs-valente:
Status check before Sprint 3 closes — has the Wallet Issuer account / GCP project been created yet? Has production publishing access been requested? No visibility on this from the issue or git history, and #3 is queued to start in Sprint 4 right behind it.

PR #363, opened 2026-09-16
docs: document Google Wallet infrastructure setup

hugo2006alm:
## Summary
- documents the external Google Wallet issuer/GCP/service-account setup required by #340
- records the intended environment/secret contract for the later #3 implementation
- documents the demo Generic Class/Object and publishing-access flow
- calls out that the current Fallstack student QR is a 30-minute JWT, so the persistent Wallet identifier must be solved in #3 rather than freezing that token into a pass

## Scope
This is intentionally documentation-only. #340 is primarily external infrastructure work; it does not require attendee-facing application code. The actual Wallet endpoint/service/UI/tests remain in #3 after the issuer setup is ready.

## Validation
Documentation checked against the current Google Wallet Generic Pass onboarding, REST credential, class/object and publishing-access documentation.

Refs #340
Refs #3


docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,215 @@
...
+
+| Variable | Secret | Purpose |
+| --- | --- | --- |
+| `GOOGLE_WALLET_ISSUER_ID` | No | Google Wallet Issuer ID. |
+| `GOOGLE_WALLET_CLASS_ID` | No | Full Generic Class ID (`issuerId.suffix`) for that environment. |
+| `GOOGLE_WALLET_SERVICE_ACCOUNT_JSON_B64` | **Yes** | Base64-encoded service-account JSON, decoded only on the server. |
+
+These variables are documented now so infrastructure can be prepared before #3,
+but they should only be added to the application's validated runtime schema when
+#3 starts consuming them. Base64 is only a transport format; it does not make the
```
david-fs-valente, line 74: This "documented now, wait for #3 to start consuming them" framing is already stale: PR #364 (same author, opened ~18 minutes after this one) already adds these exact env vars to `env.server.ts`/`.env.example` and implements the `/api/wallet` endpoint. Please reword this section (and the "Refs #340 / Refs #3" framing elsewhere in the doc) so it doesn't imply #3 hasn't started, or clarify the intended merge order between #363 and #364.

david-fs-valente (changes requested):
Doc-only change, technically accurate against the actual code and against Google's Wallet docs (verified: env var names match #364's implementation exactly, QR/JWT flow description matches src/app/api/qrcode and src/app/api/saved, external links resolve). No AGENTS.md violations (no db push guidance, no raw process.env guidance, no secret material pasted). Requesting changes for one accuracy issue before merge; the README-link nit below is optional.

docs/GOOGLE_WALLET.md
hugo2006alm: Addressed in 6ea5f7d. The guide now states that #364 implements #3, removes future-tense wording, and explicitly documents merge order #363 -> #364.

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
...
+PR #364 uses this server-only configuration contract:
+
+| Variable                                 | Secret  | Purpose                                                          |
+| ---------------------------------------- | ------- | ---------------------------------------------------------------- |
+| `GOOGLE_WALLET_ISSUER_ID`                | No      | Google Wallet Issuer ID.                                         |
+| `GOOGLE_WALLET_CLASS_ID`                 | No      | Full Generic Class ID (`issuerId.suffix`) for that environment.  |
+| `GOOGLE_WALLET_SERVICE_ACCOUNT_JSON_B64` | **Yes** | Base64-encoded service-account JSON, decoded only on the server. |
+
+PR #364 adds these variables to the application's validated runtime schema. They
+must be configured in each environment before that application change is
```
david-fs-valente: This doesn't match #364. There, all three variables are optional in `env.server.ts` ("Optional at process level so non-Wallet environments still boot"), and when they're missing `POST /api/wallet` returns 503 while the "Adicionar ao Google Wallet" button is still shown. Suggest saying they're optional at boot, but that all three must be set in any environment where the button is exposed.

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
...
+Official credential documentation:
+https://developers.google.com/wallet/generic/getting-started/auth/rest
+
+PR #364 uses this server-only configuration contract:
+
+| Variable                                 | Secret  | Purpose                                                          |
+| ---------------------------------------- | ------- | ---------------------------------------------------------------- |
+| `GOOGLE_WALLET_ISSUER_ID`                | No      | Google Wallet Issuer ID.                                         |
+| `GOOGLE_WALLET_CLASS_ID`                 | No      | Full Generic Class ID (`issuerId.suffix`) for that environment.  |
+| `GOOGLE_WALLET_SERVICE_ACCOUNT_JSON_B64` | **Yes** | Base64-encoded service-account JSON, decoded only on the server. |
```
david-fs-valente: Please add how to produce this value. #364's parser rejects anything that doesn't round-trip exactly, and GNU `base64` wraps at 76 columns by default, so `base64 key.json` pasted into `.env` or Coolify gives a 503 "credentials are invalid". Something like:

```sh
base64 -w0 key.json   # Linux
base64 -i key.json    # macOS
```

Also state that the value must be a single line of standard base64 with padding, encoding the complete JSON key file (it must contain `client_email` and `private_key`).

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
+# Google Wallet infrastructure
+
+This document covers the infrastructure/onboarding work tracked by issue #340.
+The attendee-facing Wallet integration tracked by #3 is implemented separately
+in PR #364.
+
+PR #363 (this infrastructure guide) must merge before PR #364. This keeps the
```
david-fs-valente: Merge-order and PR-number references (here, and at lines 70, 78, 158 and in the handoff section) will go stale once both PRs merge. Someone setting this up next edition needs the current contract, not the merge order. Suggest moving the merge-order note to the PR description and pointing at the code instead: `src/config/env.server.ts`, `src/application/services/googleWalletService.ts`, `POST /api/wallet`.

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
...
+3. Record the GCP project ID in ops documentation. The application does not need
+   to expose this value to the browser.
+
+## 3. Create and authorize the service account
+
+1. Create a service account in the chosen GCP project.
+2. Create a JSON key for it for the initial server-side integration.
+3. Copy the service account email.
+4. In the Google Pay & Wallet console, open **Users** and invite that service
+   account email with the **Developer** role.
```
david-fs-valente: Worth adding: grant the service account **no** IAM roles on the GCP project. Wallet access comes only from this Developer invite. People often add Editor "to make it work". It may also help to mention that the `iam.disableServiceAccountKeyCreation` org policy can block step 2, and who to ask about it.

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
...
+PR #364 adds these variables to the application's validated runtime schema. They
+must be configured in each environment before that application change is
+deployed. Base64 is only a transport format; it does not make the credential
+non-secret.
+
+## 4. Create the Fallstack Generic Class
+
+Create a **Generic** pass class in the Google Wallet Business Console.
+
+Use a separate class per hosted environment so staging changes cannot alter the
```
david-fs-valente: Classes are per environment, but #364's object IDs (`<issuer>.fallstack-2026-<studentId>`) aren't. If a DB is ever cloned between environments on the same issuer (e.g. staging restored from prod), staging would overwrite production passes. Worth a warning here, or recommend a separate issuer for non-prod.

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
...
+| `GOOGLE_WALLET_SERVICE_ACCOUNT_JSON_B64` | **Yes** | Base64-encoded service-account JSON, decoded only on the server. |
+
+PR #364 adds these variables to the application's validated runtime schema. They
+must be configured in each environment before that application change is
+deployed. Base64 is only a transport format; it does not make the credential
+non-secret.
+
+## 4. Create the Fallstack Generic Class
+
+Create a **Generic** pass class in the Google Wallet Business Console.
```
david-fs-valente: Nit: this is the only place it's called "Google Wallet Business Console". Everywhere else it's "Google Pay & Wallet console". Same console, so pick one name.

docs/GOOGLE_WALLET.md
```diff
@@ -0,0 +1,219 @@
...
+
+Do this early: approval is external to the Fallstack deployment process and PR
+#364 must not be deployed while basic issuer setup is incomplete.
+
+## 8. Environment and secret placement
+
+| Environment | Class                           | Credential location        | Notes                                                  |
+| ----------- | ------------------------------- | -------------------------- | ------------------------------------------------------ |
+| Local/dev   | Dedicated dev class             | Local untracked `.env`     | Use only test accounts; never commit the JSON key.     |
+| Staging     | Dedicated staging class         | Coolify staging secrets    | Main end-to-end validation target before release.      |
```
david-fs-valente: Please specify that these are Coolify runtime variables, not build variables. They're server-only, and shouldn't end up in image layers. Also note that the save JWT's `origins` is taken from `NEXT_PUBLIC_BASE_URL`, so that value must be the real public host in each environment.

david-fs-valente (changes requested):
Thanks, this is a clear walkthrough. The env var names and the `ISSUER_ID.SUFFIX` class format match #364, the Google links all resolve, and there are no real IDs or secrets in it.

I'm requesting changes for two accuracy problems that will trip up whoever sets this up. Both are inline:
- the doc says the variables are required, but #364 makes them optional;
- there's no base64 encoding command, and #364's parser is strict.

**Not inline:**
- **Link from README.** `README.md` links every other `docs/*.md` (OBSERVABILITY, database-workflow, SHARED_DATA). Please add a one-line pointer so this can be found.
- **Deployment caveat.** #364 sets the compose file aside: `docker-compose.app.yml` has an explicit `environment:` list and no `env_file`. Until #364 adds the three `GOOGLE_WALLET_*` variables there, setting them in Coolify as this doc says won't reach the container.

PR #364, opened 2026-09-16
feat: add attendee pass to Google Wallet

hugo2006alm:
## Summary
- adds a student-only `POST /api/wallet` endpoint that derives pass ownership exclusively from the authenticated session
- creates or patches a deterministic Google Wallet Generic Object per student through the Wallet REST API
- signs a Google `savetowallet` JWT server-side and exposes an `Adicionar ao Google Wallet` action in the existing QR/profile UI
- keeps the existing 30-minute web QR unchanged; the persistent Wallet QR contains the stable 4-character `Student.code`
- extends the company scanner so a Wallet student code is validated server-side and converted into the existing short-lived student token before continuing through `/saved`
- adds optional validated Wallet configuration and keeps service-account material server-only
- covers object creation/update/idempotency, route ownership input, the Wallet button, and student-code recognition with tests

## Security / ownership
The client never supplies a student ID/code to `/api/wallet`. The endpoint uses `session.student` only. Wallet Object IDs are deterministic (`<issuer>.fallstack-2026-<student UUID>`), and retries GET/POST/PATCH the same object rather than creating duplicates.

The Wallet barcode deliberately contains only the already-shareable `Student.code`; it does not contain the internal UUID, email, profile data, service-account material, or the expiring JWT used by the web QR.

## Configuration
Hosted environments that enable Wallet need:
- `GOOGLE_WALLET_ISSUER_ID`
- `GOOGLE_WALLET_CLASS_ID`
- `GOOGLE_WALLET_SERVICE_ACCOUNT_JSON_B64` (secret; complete service-account JSON encoded as base64)

`NEXT_PUBLIC_BASE_URL` supplies the approved origin for the signed Save-to-Wallet JWT.

## Remaining external validation
This PR implements the application side of #3. End-to-end staging validation still depends on #340's issuer/class/service-account setup and publishing access.

Refs #3
Depends on #340


src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,212 @@
...
+  return {
+    issuerId,
+    classId,
+    credentials,
+    origin: new URL(clientEnv.NEXT_PUBLIC_BASE_URL).origin,
+  };
+}
+
+export function buildGoogleWalletObjectId(issuerId: string, studentId: string) {
+  return `${issuerId}.fallstack-2026-${studentId}`;
```
david-fs-valente, line 68: `"fallstack-2026"` is hardcoded here (and `"Fallstack 2026"` at line 80) to build every Wallet object ID and card title. Per AGENTS.md's convention table, edition-specific branding belongs in `src/edition/`, not inline in a service. As written, next year's cutover requires someone to remember to grep this literal inside a service file; nothing flags it as edition-coupled. Please move both strings into `src/edition/`.

src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,212 @@
...
+function localized(value: string) {
+  return {
+    defaultValue: {
+      language: "pt-PT",
+      value,
+    },
+  };
+}
+
+function getWalletConfig() {
```
david-fs-valente, line 40: None of `getWalletConfig()`'s guard clauses (missing config, issuer/class mismatch, malformed base64/JSON credentials -> all 503) have a test, despite the PR description explicitly calling out safe-config handling. Please add coverage for each 503 branch.

src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,212 @@
...
+    method: "POST",
+    headers: walletHeaders(accessToken),
+    body: JSON.stringify(object),
+  });
+
+  if (inserted.ok) return;
+
+  // A simultaneous retry may have created the deterministic object after our
+  // GET but before our POST. Reconcile it instead of creating another object.
+  if (inserted.status === 409) {
```
david-fs-valente, line 173: This is the one branch the code comments specifically explain (concurrent request creates the object between our GET and POST, so POST returns 409 and we fall back to PATCH) and it's exactly the case the PR description highlights as intentional safety behavior — but `googleWalletService.test.ts` only covers the plain insert and plain patch paths, not this 409 fallback. Please add a test for it.

src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,212 @@
...
+    { algorithm: "RS256" }
+  );
+
+  return `https://pay.google.com/gp/v/save/${signedJwt}`;
+}
+
+export async function createGoogleWalletSaveUrl(student: WalletStudent) {
+  const { issuerId, classId, credentials, origin } = getWalletConfig();
+  const object = buildGenericObject(issuerId, classId, student);
+  const accessToken = await getAccessToken(credentials);
```
david-fs-valente, line 207: `createGoogleWalletSaveUrl` mints a brand-new OAuth assertion and exchanges it for an access token on every call via `getAccessToken`, even though the minted token is valid for up to an hour (`exp: issuedAt + 60 * 60` at line 99). Every click of "Adicionar ao Google Wallet" costs 3 outbound calls to Google instead of 2. Consider caching/reusing the token across calls within its validity window, especially under fair-day load.

src/components/QRCode/QRCodeTab/ScanTab/index.tsx
```diff
@@ -37,6 +39,39 @@ const ScanTab: React.FC<ScanTabProps> = ({ setHidden }) => {
...
+    }
+
+    setHidden(true);
+    router.push(`/student/${token}/preview`);
+    setProcessing(false);
+  }
+
+  async function handleStudentCodeScan(data: string) {
+    const code = normalizeStudentCode(data);
+    const token = await jwtStudent(code);
```
david-fs-valente, line 64: This calls the `"use server"` action `jwtStudent` directly from a client component (mirroring the pre-existing pattern in `CompanyTab`), which bypasses `lib/http/server.ts`'s `defineHandler` entirely — minting a 15-minute student-profile-preview JWT for an arbitrary 4-character code has no `auth: "employee"` gate. This PR adds a second, camera-driven entry point plus a purpose-built `isStudentCode()`/`normalizeStudentCode()` recognizer, which makes brute-forcing the ~1.68M (36^4) code space against `jwtStudent` a more obviously-fitting attack path than before. Actual student saving still requires employee auth via `/saved`, so the blast radius is "mint a preview token," not "save/exfiltrate" — but since this PR is the one widening the gap's reachability, please route this through `defineHandler` with an employee auth strategy, or otherwise gate it, rather than calling the server action unauthenticated.

src/domain/student/studentCode.ts
```diff
@@ -0,0 +1,9 @@
+const STUDENT_CODE_PATTERN = /^[A-Z0-9]{4}$/;
```
david-fs-valente, line 1: This regex re-encodes the character set/length that `src/utils/GenerateCode.ts` already defines as the source of truth for student code shape. Two independent definitions of the same format mean `isStudentCode` can silently drift from the generator and start misrouting Wallet-scanned codes in `ScanTab` if the generator format ever changes. Please derive this from the same source instead of duplicating it.

david-fs-valente (changes requested):
Core auth/security checks pass: `/api/wallet` correctly gates on `auth: "student"`, derives identity only from `session.student` (never client input), keeps Prisma access out of the service layer, reads config only through `serverEnv`, never echoes the service-account JSON back, and the object-ID idempotency logic (GET/PATCH/insert/409-fallback) is logically correct. Requesting changes on the items below before merge.

src/application/services/googleWalletService.ts
hugo2006alm: Addressed in e157e83. Wallet object prefix and card title now come from src/edition/branding.ts.

src/application/services/googleWalletService.ts
hugo2006alm: Addressed in e157e83. Tests now cover missing configuration, issuer/class mismatch, malformed base64, and malformed credential JSON, asserting each 503 response path.

src/application/services/googleWalletService.ts
hugo2006alm: Addressed in e157e83. Added the GET 404 -> POST 409 -> PATCH reconciliation regression test.

src/application/services/googleWalletService.ts
hugo2006alm: Addressed in e157e83. OAuth access tokens are cached until one minute before expires_in and concurrent token requests are deduplicated.

src/components/QRCode/QRCodeTab/ScanTab/index.tsx
hugo2006alm: Addressed in a9c1435 and 8e0144d. Preview-token minting now goes through POST /api/students/token with defineHandler auth employee and a named Zod schema; client components no longer import a server action.

src/domain/student/studentCode.ts
hugo2006alm: Addressed in 064fdea. Student code length and character set now live in domain/student/studentCode.ts and both generator and recognizer consume that source.

src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,265 @@
...
+    throw new HttpError(
+      "Google Wallet class does not belong to this issuer",
+      503
+    );
+
+  let credentials: WalletCredentials;
+  try {
+    const encoded = encodedCredentials.trim();
+    const decodedBuffer = Buffer.from(encoded, "base64");
+    if (decodedBuffer.toString("base64") !== encoded) throw new Error();
```
david-fs-valente: This round-trip check rejects the output of plain `base64 key.json`. GNU coreutils wraps its output at 76 columns by default, and `.trim()` only removes whitespace at the ends. Unpadded input fails too. An operator following the obvious path gets a 503 "credentials are invalid" with no hint why.

Suggestion: strip all whitespace before decoding (`encodedCredentials.replace(/\s+/g, "")`), or drop the round-trip check and rely on the `JSON.parse` + `serviceAccountSchema` parse to catch garbage. The docs in #363 should also say `base64 -w0` (Linux) / `base64 -i` (macOS).

src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,265 @@
...
+    method: "POST",
+    headers: { "Content-Type": "application/x-www-form-urlencoded" },
+    body: new URLSearchParams({
+      grant_type: "urn:ietf:params:oauth:grant-type:jwt-bearer",
+      assertion,
+    }),
+  });
+
+  if (!response.ok)
+    throw new HttpError("Unable to authenticate with Google Wallet", 502);
```
david-fs-valente: Operators can't see failures from Google. This `HttpError`, and the ones at lines 199, 214 and 231, drop the upstream status and body, and `httpErrorResponse` in `lib/http/server.ts` doesn't call `reportError` for `HttpError`s. So OAuth, permission, class-not-found and quota errors never reach logs or Sentry, which will make the #340 staging validation hard to debug.

Suggestion: log or `reportError` with `response.status` and a truncated body before throwing. Never log the assertion or the access token.

src/application/services/googleWalletService.ts
```diff
@@ -0,0 +1,265 @@
...
+  return {
+    issuerId,
+    classId,
+    credentials,
+    origin: new URL(clientEnv.NEXT_PUBLIC_BASE_URL).origin,
+  };
+}
+
+export function buildGoogleWalletObjectId(issuerId: string, studentId: string) {
+  return `${issuerId}.${edition.branding.wallet.objectIdPrefix}-${studentId}`;
```
david-fs-valente: This object ID has no environment segment, while the class is per environment. See the review body: if a DB is cloned between environments on the same issuer, one environment would overwrite the other's passes.

src/components/GoogleWalletButton/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const handleClick = async () => {
+    setIsLoading(true);
+    try {
+      const { url } = await getGoogleWalletSaveUrl();
+      window.open(url, "_self");
+    } catch (error) {
+      toast.error(
+        error instanceof Error
+          ? error.message
```
david-fs-valente: This shows the server's `error.message` as-is, so users see internal English config errors ("Google Wallet class does not belong to this issuer", "credentials are invalid") in an otherwise Portuguese UI. Suggest showing the Portuguese fallback for any 5xx `HttpClientError`, and only passing through messages you mean users to read.

src/components/QRCode/QRCodeTab/ProfileTab/index.tsx
```diff
@@ -108,6 +109,7 @@ const ProfileTab: React.FC<PerfilTabProps> = ({ user }) => {
           {isCopied && (
             <p className="mt-2 font-bold text-green-500">Copiado!</p>
           )}
+          <GoogleWalletButton />
```
david-fs-valente: This button renders for every student, including on environments where Wallet isn't configured (every click there returns 503) and on iOS. Consider gating it on a server-provided `walletEnabled` flag.

Also, Google's brand guidelines require the official "Add to Google Wallet" button asset rather than a custom black button: https://developers.google.com/wallet/generic/resources/brand-guidelines

src/app/api/students/token/route.ts
```diff
@@ -0,0 +1,17 @@
+import { NextResponse } from "next/server";
+
+import { defineHandler } from "@/lib/http/server";
+import { jwtStudent } from "@/application/services/studentTokenService";
+import { studentTokenSchema } from "@/schemas/studentTokenSchema";
+
+export const POST = defineHandler({
+  auth: "employee",
```
david-fs-valente: Moving the token behind `auth: "employee"` is a good fix. There's no rate limiting here, though, so an employee account could enumerate the ~1.68M 4-character codes. Fine as a follow-up, but worth tracking.

david-fs-valente (changes requested):
Thanks for this. The service design is solid: the student comes only from the session, object IDs are deterministic, the 409 race falls back to PATCH, `id`/`classId` stay out of the PATCH body, the class ID is checked against the issuer, and the credentials stay server-only. Removing `"use server"` from `studentTokenService.ts` also closes a real hole on `dev`: `jwtStudent(code)` was callable as a server action without a login.

I'm requesting changes mainly because of the first item below. As the PR stands, the feature can't be turned on in the Coolify deployment.

**Blocking: the Wallet env vars never reach the deployed container** (`docker-compose.app.yml`, which this PR doesn't change, so I can't comment inline)

The `web` service's `environment:` block (around lines 35-66) is an explicit list, and there's no `env_file`. None of `GOOGLE_WALLET_ISSUER_ID`, `GOOGLE_WALLET_CLASS_ID` or `GOOGLE_WALLET_SERVICE_ACCOUNT_JSON_B64` are in it. So even with the values set in Coolify, `getWalletConfig()` gets `undefined` and every click returns 503 "Google Wallet is not configured". Please add all three as `${VAR:-}` to the `web` environment. It would also be worth a test next to `tests/e2e/dockerBuildArgs.test.ts` that checks every `serverEnvSchema` key appears in the compose runtime env, so this can't happen again.

**Also not inline: env-scoped object IDs.** Classes are per environment, but object IDs (`<issuer>.fallstack-2026-<studentId>`) aren't. If a DB is ever cloned between environments that share the issuer (e.g. staging restored from prod), staging would PATCH production passes. Consider adding an environment segment to the prefix, or document that the issuer must not be shared in that case.

**Informational / possible follow-ups:**
- The Wallet QR holds the permanent student code, so whoever holds it can get a preview token with no 30-minute expiry. The PR description says this is intended, and it matches what manual code entry already allows. `/api/students/token` has no rate limit, though, and 36^4 is about 1.68M codes, so a malicious employee account could enumerate students. Rate limiting or audit logging would be good to add later.
- The pass has no `validTimeInterval`, and nothing expires or updates it when an account is deleted, so the student's name and code stay in Google Wallet. Setting the interval to the event dates and expiring the object on account deletion would tidy that up.

Inline comments below cover the rest.

PR #366, opened 2026-09-19
test(e2e): expand event coverage across roles

### LINKED ISSUE #341: [Test] Expand event E2E coverage to employee and admin critical flows

dinisjcorreia:
Closes #341

## Summary

- cover Student QR -> Employee scan against real staging HTTP/database state
- assert exact duplicate/retry rejection without duplicate or overwrite
- cover isolated Admin FAQ create/read/delete
- verify Student, Employee, Admin, and Super Admin role boundaries
- preserve configured Admin/Super Admin tiers while requiring AuthNEI admin grant
- document synthetic staging accounts, credentials, cleanup, and rerun safety

## Review follow-up

Supersedes #357 after its noncompliant `test/` branch was renamed to `fix/`. The event scan reads initial staging state and asserts exact `201` for a clean Student/company pair or exact `409` for a rerun. Review fixes preserve stored admin tiers through grant-less logins, document the mandatory Super Admin rollout check, and harden staging cleanup.

## Verification

- `pnpm lint`
- `pnpm typecheck`
- `TZ=UTC pnpm test` (77 files, 314 tests)
- Playwright harness: 3 expected skips without staging storage-state credentials

Live staging execution still requires the four synthetic account storage-state files described in `tests/e2e/README.md`.


src/application/repositories/userRepository.ts
```diff
@@ -72,14 +73,16 @@ export async function provisionZitadelUser(input: {
...
   if (existingBySubject) {
     return prisma.user.update({
       where: { id: existingBySubject.id },
       data: {
         email: input.email,
-        adminRole,
+        adminRole: resolveAdminRole(
+          input.isGlobalAdmin,
+          existingBySubject.adminRole
+        ),
```
david-fs-valente: **High — the configured Admin tier is not durable here, and losing it silently escalates ADMIN to SUPER_ADMIN.**

`provisionZitadelUser` runs on every OIDC login and writes `resolveAdminRole(input.isGlobalAdmin, ...)` back to `User.adminRole` in all three branches (this one, line 112, and line 128). That value is `null` whenever the ZITADEL grant is absent. Since this column is now the sole record of the tier, any single grant-less login wipes it, and the new `?? "SUPER_ADMIN"` default in `resolveAdminRole` then promotes the user to the highest tier on the next granted login.

Concrete path:

1. A super admin creates Bob in the backoffice with the form's default tier `ADMIN` (`src/components/AdminAccountForm/index.tsx:90`).
2. Bob logs in before the NEI Global grant is applied, so his row is set to `adminRole: null`.
3. The grant is applied.
4. Bob logs in again. `resolveAdminRole(true, null)` returns `SUPER_ADMIN`, and he can now create and delete admin accounts.

The same happens to any existing ADMIN whose grant is briefly revoked, or whose role claim is missing from a single token.

Suggested fix: only write `adminRole` during provisioning when `isGlobalAdmin` is true, and leave the stored tier untouched otherwise. The session layer already refuses to honor a stored tier without the grant, so nothing is escalated by preserving it.

src/application/services/sessionService.ts
```diff
@@ -20,10 +21,9 @@ const getServerSession = cache(async () => {
     const appUser = await findUserSessionByZitadelUserId(claims.sub);
     if (!appUser || !appUser.active) return null;
 
-    // ZITADEL is authoritative for privileged roles. The DB fields remain
-    // useful for domain/profile state, but stale local role data can never
-    // manufacture admin/employee authorization by itself.
-    const adminRole: "SUPER_ADMIN" | null = claims.admin ? "SUPER_ADMIN" : null;
+    // ZITADEL grant gates admin access; local tier only narrows that grant.
+    // Missing tier defaults existing global admins to Super Admin.
+    const adminRole = resolveAdminRole(claims.admin, appUser.adminRole);
```
david-fs-valente, line 26: **Medium — deploy-time lockout risk from the tier split, with no backfill or pre-deploy check.**

Before this change, every ZITADEL global admin resolved to `SUPER_ADMIN`. After it, a row whose `adminRole` is `"ADMIN"` narrows them. Every admin-management surface is `auth: "superadmin"` (`src/app/api/admin/admins/route.ts`, `src/app/api/admin/admins/[id]/route.ts`, `src/app/(admin)/admins/page.tsx`).

If the current production rows are all `"ADMIN"` — plausible, since the new-admin form defaults to `ADMIN` — then after deploy nobody resolves as SUPER_ADMIN, and the only endpoint that could promote someone itself requires SUPER_ADMIN. Admin-account management would be unrecoverable without direct database access.

Could we add either a migration/backfill, or a documented pre-deploy check that at least one active `adminRole = 'SUPER_ADMIN'` row exists?

tests/e2e/playwright/event-role-flows.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+      );
+      expect(list.status()).toBe(200);
+      await expect(list.json()).resolves.toMatchObject({
+        items: [expect.objectContaining({ id: faqId })],
+        totalCount: 1,
+      });
+    } finally {
+      if (faqId) {
+        const cleanup = await admin.delete(`/api/admin/faqs/${faqId}`);
+        expect(cleanup.status()).toBe(204);
```
david-fs-valente: **Low — `expect()` inside `finally` masks the real failure and leaks the context.**

If the `try` block fails (for example on the `totalCount: 1` assertion), the `finally` still runs the DELETE. Should cleanup return anything other than 204, this `expect` throws and *replaces* the original assertion error, so the report points at cleanup rather than at the actual regression. It also throws before `await admin.dispose()` on line 153, leaking the API context.

Prefer asserting cleanup outside `finally`, or recording it as a soft assertion (`expect.soft`) / logging it.

tests/e2e/README.md
```diff
@@ -33,9 +34,15 @@ pnpm exec playwright codegen \
   https://staging.example.org/login
 ```
 
-Repeat that command whenever the session expires. Delete the file immediately
-after testing (`rm tests/e2e/.staging-student.json`; PowerShell:
-`Remove-Item tests/e2e/.staging-student.json`).
+Repeat this command for `.staging-employee.json`, `.staging-admin.json`, and
+`.staging-super-admin.json`, signing in as the matching synthetic account each
+time. The Employee account must belong to a synthetic company. Verify the Admin
+account resolves as `ADMIN`, not `SUPER_ADMIN`; the suite checks every supplied
```
david-fs-valente: **Low — these setup instructions can't produce the ADMIN-tier account the suite requires.**

The README says to capture `.staging-admin.json` by signing in, and to "Verify the Admin account resolves as `ADMIN`, not `SUPER_ADMIN`" — but gives no way to make that true. Per `resolveAdminRole`, a freshly granted NEI Global admin with no pre-existing `adminRole` row resolves to `SUPER_ADMIN`.

Following the README literally yields two SUPER_ADMIN accounts, and then both `expectSession(admin, { adminRole: "ADMIN" })` (spec lines 130 and 175) and the `admin.get("/api/admin/admins")` 403 assertion (spec line 183) fail for reasons unrelated to the code under test.

Please document the required step: create the synthetic Admin through the backoffice with tier `ADMIN` (or set `User.adminRole = 'ADMIN'` on staging) *before* its first login.

david-fs-valente (changes requested):
Reviewed the ADMIN/SUPER_ADMIN split and the new cross-role Playwright suite. The e2e flow logic itself checks out against the real routes, but the tier change that the suite depends on has a privilege-escalation path and a deploy-time lockout risk, so I'd like those addressed before this merges.

Verified locally in a throwaway worktree: `eslint` clean on all changed files, and `sessionService.test.ts` + `authPolicy.test.ts` pass (16 tests). `vitest.config.ts` excludes `tests/e2e/playwright/**`, so the new spec correctly stays out of `pnpm test`.

**Summary of findings**

1. High — `provisionZitadelUser` still overwrites `User.adminRole` on every login, so the stored tier is not durable and a grant-less login can silently escalate an ADMIN to SUPER_ADMIN.
2. Medium — no backfill or pre-deploy check that at least one `SUPER_ADMIN` row exists; if production rows are all `ADMIN`, admin-account management becomes unrecoverable after deploy.
3. Low — `expect()` inside `finally` in the FAQ test masks the real failure and leaks the API context.
4. Low — the README setup steps cannot produce the ADMIN-tier account the suite asserts on.

Out of scope, but worth knowing: `/api/students/[code]/history` returns the saving company's private `comment` to the student, and the new test now locks that exposure in.

src/application/repositories/userRepository.ts
dinisjcorreia: Valid. Fixed in 6eb7727. Grant-less provisioning now returns no stored-tier update, so both existing-user update paths omit `adminRole` entirely. This preserves `ADMIN`/`SUPER_ADMIN` through missing or revoked claims and avoids a stale-read overwrite race. Session authorization still returns `null` without the external grant. Added focused regression coverage.

src/application/services/sessionService.ts
dinisjcorreia: Valid. Added an Admin-tier rollout pre-check to `docs/database-workflow.md` in 6eb7727: exact query for an active `SUPER_ADMIN`, manual promotion SQL when none exists, mandatory re-check, and explicit stop-deployment condition. SQL was documented only and not executed.

tests/e2e/playwright/event-role-flows.spec.ts
dinisjcorreia: Valid. Fixed in 00334d0. Cleanup now uses `expect.soft`, and nested `try/finally` guarantees `admin.dispose()` even if DELETE/request cleanup fails. Original test failure remains visible.

tests/e2e/README.md
dinisjcorreia: Valid. Fixed in 00334d0. README now explains the required sequence: Super Admin creates the synthetic account locally with tier `ADMIN`, then its NEI Global grant is applied, then storage state is captured. It also warns that a fresh grant without a configured tier defaults to `SUPER_ADMIN`.

dinisjcorreia:
Addressed all four review findings in `6eb7727` and `00334d0`.

- preserved stored admin tier when the external claim is absent
- documented mandatory pre-deploy Super Admin check/manual promotion
- made FAQ cleanup non-masking and disposal unconditional
- documented how to provision a true synthetic `ADMIN` account

Full pre-push gate passed: lint, typecheck, 77 test files / 314 tests. Live staging flow remains pending credentials. Please re-review.

tests/e2e/playwright/event-role-flows.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      await expectSession(employee, { role: "EMPLOYEE", adminRole: null });
+      await expectSession(admin, { adminRole: "ADMIN" });
+      await expectSession(superAdmin, { adminRole: "SUPER_ADMIN" });
+
+      expect((await anonymous.post("/api/saved")).status()).toBe(401);
+      expect((await student.post("/api/saved")).status()).toBe(403);
+      expect((await student.get("/api/admin/faqs")).status()).toBe(403);
+      expect((await employee.get("/api/qrcode")).status()).toBe(403);
+      expect((await employee.get("/api/admin/faqs")).status()).toBe(403);
+      expect((await admin.get("/api/admin/admins")).status()).toBe(403);
```
david-fs-valente, line 187: This block asserts the negative case for ADMIN (blocked from the super-admin-only `/api/admin/admins`) but never the positive one. Every other role here gets both sides: STUDENT and EMPLOYEE are checked against endpoints they may not reach, and SUPER_ADMIN is checked for 200 on both. Without a positive assertion, a regression that locked ADMIN out of the ordinary admin surface entirely would still pass this test.

Suggest adding, alongside the existing line:

```ts
expect((await admin.get("/api/admin/faqs")).status()).toBe(200);
```

Non-blocking.

david-fs-valente (approved):
Reviewed the expanded role coverage. The tests are solid and the accompanying fix is a real improvement, so approving — two non-blocking notes below.

**Test quality** — the new unit tests exercise real logic rather than mirroring mocks: `authPolicy.test.ts` tests `resolveAdminRole` and `resolveStoredAdminRole` as pure functions, and `sessionService.test.ts` checks that a locally configured tier survives behind the ZITADEL grant gate. No wall-clock or date dependence, no shared mutable state between tests, and the Playwright specs clean up in `finally` and tolerate reruns (they read initial state and accept both 201 and 409).

**Gating** — the Playwright suite is skipped unless both `E2E_BASE_URL` and `CONFIRM_NON_PRODUCTION` are set, so CI will not reach a real environment by accident. Placement matches the conventions in AGENTS.md: unit tests colocated, integration specs under `tests/e2e/playwright/`, shared helper in `tests/e2e/roleTestUtils.ts`.

**Production change** — this PR is titled `test(e2e):` but it also changes behaviour: `authPolicy.ts` gains `resolveAdminRole`/`resolveStoredAdminRole`, and `sessionService.ts` and `userRepository.ts` now route through them. That fixes a genuine bug (an ADMIN-tier user was reset to SUPER_ADMIN or null on every re-provisioning; the tier is now preserved unless the ZITADEL grant is revoked), and returning `undefined` to skip the write when there is no grant is the right shape. My only note is the labelling: AGENTS.md asks for unrelated concerns to be split into separately scoped commits, and a behaviour fix of this size reads better as its own `fix:` commit than folded under a `test(e2e):` title. Worth fixing the PR title at minimum, since the changelog for the edition is hand-written from these.

**Docs** — the additions to `.env.example`, `tests/e2e/README.md`, and the SUPER_ADMIN pre-deployment check in `docs/database-workflow.md` are welcome and match what the suite actually needs.

CI is green on all eight checks.

tests/e2e/playwright/event-role-flows.spec.ts
dinisjcorreia: Valid. Added the positive `ADMIN -> GET /api/admin/faqs -> 200` assertion in d452d18, alongside the existing super-admin-only denial. This now proves both sides of the ADMIN boundary.

dinisjcorreia:
Addressed both non-blocking review notes.

- `d452d18` adds positive ordinary-admin access coverage alongside the super-admin-only denial.
- PR title now uses `fix(auth):` to reflect the production auth behavior change.

Focused lint, typecheck, and Playwright harness checks passed. Full suite skipped for this follow-up at requester direction. Please re-review.

PR #373, opened 2026-09-20
refactor: split logo storage into dedicated bucket

### LINKED ISSUE #283: [Refactor] Split logo storage into a dedicated Supabase bucket
### Description
Separate company/sponsor logo storage from student avatars so logo policy can evolve independently without weakening the avatar pipeline.

### Why
Company/sponsor logos are currently uploaded through the same `admin/storage/avatar` route into the same `avatars` Supabase bucket as student self-service avatar uploads, under the same type/size policy (`config.uploads.avatar`). Official, admin-curated brand assets and self-service personal photos are different trust surfaces: one is uploaded by a small set of admins and rendered as site chrome, the other is uploaded by any student and rendered as user content. They need independent allowed types, size limits, path conventions and bucket policy. Sharing one bucket and one config means neither can change without weighing the risk to the other.

Vector logo support (#282) is one future consumer of that separation, and it is deferred beyond the 2026 edition. This issue does not depend on it and does not wait for it: the decoupling is worth doing on its own.

### Scope
- Create a dedicated public `logos` Supabase Storage bucket.
- Add a logo-specific upload configuration (types, limits, path conventions) as its own config object, not a reference to the avatar config.
- The logos bucket accepts the current raster set (webp/png) at launch. SVG is **not** enabled by this issue.
- Point company/sponsor logo uploads at the new bucket and rename/split the admin upload route if needed for clarity.
- Update setup documentation and bucket-limit/policy SQL to make the environment reproducible.
- Add tests for bucket selection, permissions and upload validation.

### Migration decision
New uploads only. Existing logo objects are **not** migrated bucket-to-bucket; the current logo set is re-uploaded manually through the admin backoffice after the new pipeline lands. Re-uploading a handful of logos is a few minutes of admin work with no data-migration risk, while a bucket-to-bucket copy with stored-URL rewrites is a real migration running weeks before the event.

### Scheduling constraint
Hard cutoff **2026-11-01**. This issue touches company/sponsor logo rendering, which is visible on the home page. If it has not merged by the close of the readiness window, it moves past the event together with #282. It must not be in flight when the release-readiness gate (#344) begins.

### Acceptance Criteria
- A dedicated public `logos` Supabase Storage bucket exists, separate from the avatars bucket.
- Logo uploads use their own configuration object (allowed types, size limits, path convention), independent of the avatar upload configuration — even though both currently permit the same raster formats.
- Company/sponsor logo uploads write to the new `logos` bucket instead of the avatars bucket.
- Student avatar uploads are unaffected and continue to use their own bucket and configuration.
- Setup documentation and bucket-limit/policy SQL are updated so the environment is reproducible from a clean setup.
- The new-uploads-only migration decision is reflected in the setup documentation, including the manual re-upload step.
- Tests cover bucket selection, permissions, and upload validation for the new logo pipeline.

### Dependencies
- [ ] Coordinate with #284 — legacy logo asset removal happens after this pipeline exists.

dinisjcorreia:
Closes #283

## Summary

- add an admin-only logo upload route backed by the public `logos` bucket
- give logos independent bucket, path, MIME, and size configuration while leaving student avatars unchanged
- validate PNG/WebP signatures and route company/sponsor forms through the logo pipeline
- make bucket creation/policy reproducible and document the new-uploads-only manual re-upload

## Verification

- `pnpm lint` (pre-existing warnings only)
- `pnpm exec vitest run --exclude 'tests/e2e/**'` — 307 passed
- `pnpm typecheck`
- `pnpm build`
- `pnpm exec prisma validate`
- dev smoke: unauthenticated multipart POST to `/api/admin/storage/logo` returned 403

DB-backed integration tests were not run locally because no test `DATABASE_URL` is available in this worktree; CI remains authoritative for them.


david-fs-valente (approved):
Reviewed the logo storage split. No correctness issues found.

**Bucket semantics** — `logos` is created public, matching `avatars`, while `cvs` stays private. The bucket SQL uses `on conflict do update`, so it is safe to re-run and creates the bucket where it is missing.

**Back-compat** — existing objects stay readable in `avatars`, `logoSchema` accepts both the old `avatars` URLs and the new `logos` URLs, and the README documents the manual re-upload path. Nothing breaks for records still pointing at the old bucket.

**Migrations and env** — no Prisma schema change, so no migration is needed, and no new env vars were introduced (configuration stays in `src/config/index.ts`), so nothing to sync into `.env.example` or the Docker build args.

**Upload route** — admin-only session check, file type and size limits enforced from config, file-signature validation, and UUID object keys so keys are not enumerable. The multipart exception to `defineHandler` is documented in a route comment, consistent with the existing `storage/avatar` and `storage/cv` routes.

**Responses** — every failure path returns `{ error: string }` with an explicit status (400 validation, 403 auth, 500 server), and creation returns 201 rather than relying on the implicit 200 the older avatar route falls back to.

**Signature check** — the refactor from `some` to `every` is the right fix for multi-byte signatures; the WebP pattern (RIFF at offset 0, WEBP at offset 8) only validates correctly under `every`.

**Tests** — five cases cover auth rejection, a valid WebP upload, JPEG rejection, signature mismatch, and the size limit. CI is green.

Approving.

PR #374, opened 2026-09-20
chore: remove legacy company logo assets and fallbacks

### LINKED ISSUE #284: [Chore] Remove dead legacy company logo assets
### Description
Remove the legacy static company-logo assets and fallback rendering paths that predate the current DB-backed company/logo model.

### Scope
- Confirm the legacy `.webp` tier directories and old `CompanyImage` fallback paths are no longer required by current or migration code.
- Remove dead logo files and unreachable fallback code.
- Verify company/sponsor rendering uses the current DB/storage pipeline after the #283 bucket split.
- Add/update any small regression coverage needed to prevent fallback paths from silently returning.

### Acceptance Criteria
- No references to the legacy `.webp` tier directories or `CompanyImage` fallback paths remain in application or migration code.
- The legacy asset files are deleted from the repository.
- Company/sponsor logos render correctly end-to-end using only the current DB/storage pipeline, with no fallback path exercised.
- Regression coverage asserts rendering does not silently fall back to the removed legacy path.

### Dependencies
- [ ] #283 — the dedicated logo pipeline must exist, and its manual logo re-upload must be done, before the legacy fallback path is removed.

dinisjcorreia:
Closes #284

Depends on #373. Do not merge until #373 is deployed and every current company/sponsor logo has been manually re-uploaded through the admin backoffice.

## Summary

- delete 22 legacy company/sponsor logo assets
- remove `CompanyImage` static fallback behavior and render only stored logo URLs
- reject legacy local logo paths in admin validation
- remove legacy asset seed references from clean-database migration history
- add regression coverage for stored-logo rendering and absent-logo behavior

Existing deployed databases are unchanged by this PR; removed seed blocks only affect fresh database builds.

## Verification

- `pnpm lint` (pre-existing warnings only)
- `pnpm exec vitest run --exclude 'tests/e2e/**'` — 304 passed
- `pnpm typecheck`
- `pnpm build`
- `pnpm exec prisma validate`

DB-backed integration tests were not run locally because no test `DATABASE_URL` is available in this worktree; CI remains authoritative for them.


prisma/migrations/20260714011113_company_display_fields/migration.sql
```diff
@@ -1,36 +1,5 @@
...
-ON CONFLICT ("name") DO UPDATE SET
-  "tier" = EXCLUDED."tier",
-  "avatar" = EXCLUDED."avatar",
-  "website" = EXCLUDED."website",
-  "active" = EXCLUDED."active",
-  "order" = EXCLUDED."order";
+ALTER TABLE "Company"
+ADD COLUMN "active" BOOLEAN NOT NULL DEFAULT false,
+ADD COLUMN "order" INTEGER NOT NULL DEFAULT 0,
+ADD COLUMN "website" TEXT;
```
david-fs-valente: **Blocker.** This migration has already been applied to dev, staging and production, so editing it breaks its recorded checksum in `_prisma_migrations` and `prisma migrate deploy` will abort with "The migration `20260714011113_company_display_fields` was modified after it was applied." Because the migrator service runs before the app is marked healthy, that fails the whole deploy rather than degrading quietly.

The `ALTER TABLE` reformatting in this hunk is caught by the same rule — any byte change to an applied migration invalidates the checksum, cosmetic or not.

Restore this file exactly as it was and move the backfill removal into a new forward migration that deletes or rewrites the seeded rows.

prisma/migrations/20260714023514_add_sponsor_table/migration.sql
```diff
@@ -14,17 +14,3 @@ CREATE TABLE "Sponsor" (
 
 -- CreateIndex
 CREATE UNIQUE INDEX "Sponsor_name_key" ON "Sponsor"("name");
```
david-fs-valente, line 16: **Blocker,** same checksum problem as `20260714011113_company_display_fields`: this migration is already applied, so removing the sponsor backfill from it fails `migrate deploy`.

It also does not achieve the goal on its own. Every environment that already ran this migration still has the seven sponsor rows pointing at `/assets/images/sponsors/*.png`, and those files are deleted in this PR.

Note for the replacement migration: `Sponsor.logo` is `String`, not `String?`, so these rows cannot simply be nulled the way `Company.avatar` can. They need a placeholder value, a re-upload, or deletion.

src/schemas/logoSchema.ts
```diff
@@ -69,22 +69,8 @@ function isAllowedLogoUrl(value: string): boolean {
...
-      (value.startsWith("/") && !value.startsWith("//")) ||
-      isAllowedLogoUrl(value),
-    {
-      message:
-        'Must be an absolute path starting with "/" or a URL from configured storage',
-    }
-  );
+// Logos must come from configured remote Storage. Legacy public/ asset paths
+// are deliberately rejected now that company/sponsor logos use Storage only.
+export const logoSchema = z.string().max(2048).refine(isAllowedLogoUrl, {
```
david-fs-valente, line 74: Tightening this to Storage-only is the right call, but it needs the data cleanup to land first or in the same deploy. Any company or sponsor row still holding `/assets/images/...` now fails validation, so an admin opening that record to replace the dead logo cannot save it — the record is simultaneously rendering a 404 and rejecting the edit that would fix it.

No change needed to this file itself; sequencing it behind the forward migration resolves it.

david-fs-valente (changes requested):
Right direction — moving company and sponsor logos to Storage and dropping the static assets is the correct end state, and the dead-reference cleanup is thorough. I checked every deleted asset against the tree and the only surviving mention of `/assets/images/companies` or `/assets/images/sponsors` is the negative assertion in `src/schemas/logoSchema.test.ts:19`, with no dynamically built paths left behind. The dependency on #373 is satisfied, so merge order is not a concern.

Requesting changes on one blocker and one data-integrity gap.

**Blocker: two already-applied migrations are edited in place.** Prisma records a checksum per migration in `_prisma_migrations`, so rewriting a migration that has already run makes `migrate deploy` fail with "was modified after it was applied" rather than skipping it. AGENTS.md has migrations deploying through the migrator service before the application becomes healthy, so this fails the deploy on every environment that already ran these two — dev, staging and production. The fix is to restore both files byte-for-byte and put the cleanup in a new forward migration.

**Data gap: deployed rows still point at the deleted files.** Removing the backfill only changes what a fresh database gets. Existing rows keep their `/assets/images/...` values, and those files are gone, so the public pages render 404s. The tightened `logoSchema` compounds it: those values no longer validate, so an admin who opens such a record to fix the logo cannot save it either. The record ends up both broken and uneditable.

Worth noting for whichever shape the forward migration takes: `Company.avatar` is `String?`, but `Sponsor.logo` is `String` and not nullable, so nulling out the stale paths only works for companies. Sponsors need a real replacement value, a placeholder, or row deletion, and either way the re-upload step should be written down the way #373 documented its own.

The `CompanyImage` simplification itself is fine — `if (!company.avatar?.trim()) return null;` yields an empty slot rather than a broken-image icon. It just does not help here, since a stale path is truthy and still renders.

Left as draft, which matches: the migration handling looks like the unfinished part rather than an oversight.

prisma/migrations/20260714011113_company_display_fields/migration.sql
dinisjcorreia: Fixed in 5a98aba. Restored  byte-for-byte and verified its Git blob hash matches . Legacy company-logo cleanup now lives in the new forward migration.

prisma/migrations/20260714023514_add_sponsor_table/migration.sql
dinisjcorreia: Fixed in 5a98aba. Restored  byte-for-byte. The forward migration deletes only sponsor rows still using removed  paths; rows already using Supabase Storage URLs remain unchanged.

src/schemas/logoSchema.ts
dinisjcorreia: Fixed in 5a98aba. The new forward migration clears stale company paths and removes stale sponsor rows in the same deploy. Production runs Prisma migrations before application health, so storage-only validation cannot serve those legacy references.

dinisjcorreia:
@david-fs-valente All requested changes are addressed in 5a98aba, with each review thread replied to and resolved. Please re-review when available.

prisma/migrations/20260921190000_remove_legacy_logo_references/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- Preserve storage-backed logos while removing references to deleted static assets.
+UPDATE "Company"
+SET "avatar" = NULL
+WHERE "avatar" LIKE '/assets/images/companies/%';
+
+-- Sponsor.logo is required, so stale legacy rows cannot remain editable or renderable.
+-- Current sponsors re-uploaded through the admin backoffice use Storage URLs and remain.
+DELETE FROM "Sponsor"
```
david-fs-valente: This permanently deletes the name, website and order of every sponsor that wasn't re-uploaded before this deploys. The migrator runs on deploy, so a merge before the production re-upload is finished destroys data with no rollback.

Please do one of these:
- Confirm, with a query result, that production has zero rows matching `logo LIKE '/assets/images/sponsors/%'` before merging.
- Make this reversible, e.g. `UPDATE "Sponsor" SET "active" = false WHERE ...`, so an admin can re-upload later. The admin `LogoThumbnail` would show a broken image for those rows until then, but editing still works.

prisma/migrations/20260921190000_remove_legacy_logo_references/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- Preserve storage-backed logos while removing references to deleted static assets.
+UPDATE "Company"
+SET "avatar" = NULL
+WHERE "avatar" LIKE '/assets/images/companies/%';
```
david-fs-valente: This clears the avatar, and `CompaniesSection` then drops these companies from the public roster entirely (see the review body).

src/schemas/logoSchema.ts
david-fs-valente: `CompanyForm` and `SponsorForm` resubmit the current `avatar`/`logo` on every save. The old schema accepted any `/...` path, so an admin could have saved one outside the two prefixes the migration covers. Any such row would now fail validation on unrelated edits until its logo is replaced.

Suggestion: check production for `avatar LIKE '/%'` and `logo LIKE '/%'` before deploy, or widen the migration's `WHERE` to `LIKE '/%' AND NOT LIKE '//%'`.

david-fs-valente (changes requested):
The code side looks clean. I found no remaining references to the deleted files in `src/`, `edition/`, `prisma/seed.ts` or config; the only hits are the historical backfill migrations. `CompanyImage` is simpler, and the other logo consumers already handle a missing logo.

I'm requesting changes because of the data migration. It runs automatically on deploy and can't be rolled back, and right now the only safeguard is a line in the PR description.

**Not inline: companies without a logo disappear from the public roster** (`src/components/Companies/CompaniesSection/index.tsx:15`, which this PR doesn't change)

`toCompanyProps` returns `null` when `avatar` is empty. So every company this migration clears is removed from the public companies section entirely. The PR body says they're only shown "with no logo". Please either say this in the PR and deploy checklist, or render a name-only card for companies without a logo.

**Fresh environments.** On any new DB (local `pnpm migrate`, staging, CI), the backfill migrations insert the legacy rows and this migration then deletes or clears them. `prisma/seed.ts` sets no avatars or sponsors, so local and staging end up with no sponsors and no visible companies. Worth a note in the README/PR, or seeding placeholder Storage logos.

**Staging check.** AGENTS.md asks for a repeatable staging check when a test isn't practical. Please add before/after `SELECT count(*)` queries for both tables to the PR body.

